PR #4, opened 2018-05-23
Add Rubocop with initial configuration

tibbon:
This adds Rubocop without any other code changes to update styles. 

Suggested path forward on conforming to the guideguide is to attempt to remove a single style at a time from the `.rubocop_todo.yml`, and then running `rubocop --auto-correct` if it's a style that supports autocorrect, or manually making the changes and then making a single commit/PR for updating that style. 

Alternately, if it's an intentional style that doesn't conform to the defaults, remove it from the `.rubocop_todo.yml` file and put an intentional exclusion with comment in `.rubocop.yml`.


gamafranco:
I suspect this will raise the memory footprint of the gem in production. Any chance to make this load only in the development environment?

tibbon:
I could put it in the Gemfile in a `development` group. Does that differ from putting it in the `.gemspec` file as `s.add_development_dependency 'rubocop'`?

ddnexus:
@gamafranco: The way @tibbon add the rubocop gem is actually (almost) the less invasive way you can think of. Unless you install the gem with the `--development` flag `rubocop` would not be listed as a dependency, hence it will not even be installed with  `pagy`. 

Another alternative would be not adding it even as a development gem. I personally add the development gems to the `gemspec` for easing the installation for other people, but I install them manually without the `--development` flag during the development.

ddnexus:
@tibbon Thank you for your pull request. Rubocop could be a nice addition to keep me straight.

I am willing to merge your pull request... IF... you can configure it in such a way it will never bother me with the indentation style of the whole `Pagy` module namespace.

I don't care about any other file, like tests, and other accessories... Those files are similar to how you should use natural language in "legal-contracts", and it should be written and indented in the most conventional and clear way for the benefit of everybody, so the method for doing so - that you describe in the first post - is OK with me.

But the core working code is more like "poetry", where breaking the conventional rules may add more meaning, and it certainly does for me. And like poetry, that may not give the same vibes to everybody: it may look beautiful to someone and horrible to others, but nobody would question where the author likes to put his commas. ;)

So, please, could you rewrite the config file in order to totally exclude the `Pagy` module from every indentation style suggestions? That should remove all its files from the other auto-generated suggestions in the config, so leaving only the stuff related to the other files. As I said I am fine with those and I will try to adjust to that ASAP. 

Thanks.

grosser:
maybe remove the todo and disable all metrics/style things ?
... then can make PRs to opt-in to thing that seem worthwhile

bquorning:
I think this is may of those repositories where you disable all cops by default and then only enable the ones that should be applied:
```
AllCops:
  DisabledByDefault: true
```

See https://rubocop.readthedocs.io/en/latest/configuration/#generic-configuration-parameters

ddnexus:
@grosser & @bquorning: I like the opt-in approach! Disable all and add what is worthwhile.

grosser:
Also need:
```yaml
AllCops:
  DisplayCopNames: true
```

gamafranco:
May I suggest a pull request? Your idea will be easy to materialise if the suggestion comes directly in readme file.

Just my 5 cents.

grosser:
https://github.com/ddnexus/pagy/pull/18

grosser:
also can you enable travis https://travis-ci.org/ddnexus/pagy would be nice to know that the tests actually run before merging :D

gamafranco:
Agree. I would also but bundler-audit and brakemen, failing when a vulnerability is detected. I can do that PR once this one is merged to master.

ddnexus:
Implemented in #18

PR #7, opened 2018-05-25
Create CODE_OF_CONDUCT.md

gamafranco:
Adding code of conduct.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
```
grosser: hmm listing these things means other things are ok ? ... so I can make fun of people because they use vim ? ... just remove the list ?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
...
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable behavior and are expected to take appropriate and fair corrective action in response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, or to ban temporarily or permanently any contributor for other behaviors that they deem inappropriate, threatening, offensive, or harmful.
```
grosser: replace the list with just "contributions" ?

CODE_OF_CONDUCT.md
gamafranco: Yes. Anything else besides EMACS should result in public shaming.

ddnexus:
Hmmm.... do we really need all that? :)

Maybe it's because I have some little problem with authority, but my ideal code of conduct would be something like: 

`Please, be nice and use common sense. Thanks.`

Is that too simple? :)

grosser:
I like that :)

gamafranco:
We can go for that either. The problem with common sense is that the common denominator is sometimes too small...

But we can talk again in a few weeks when someone from Gender Neutral movement submits a pull request :)

ddnexus:
@gamafranco  LOL

PR #10, opened 2018-05-25
Remove a couple of Range allocations

bquorning:
This gem seems to be (proudly) optimized for speed and few object allocations, so I figure it’s ok to open a PR for removing even a few superfluous `Range` instantiations :-)

### DIFF
--- a/lib/pagy.rb
+++ b/lib/pagy.rb
@@ -28,7 +28,7 @@ class Pagy ; VERSION = '0.7.0'
       instance_variable_set(:"@#{k}", @vars.delete(k).to_i)               # set instance variables
     end
     @pages  = @last = [(@count.to_f / @items).ceil, 1].max                # cardinal and ordinal meanings
-    (1..@last).cover?(@page) || raise(OutOfRangeError, "expected :page in 1..#{@last}; got #{@page.inspect}")
+    @page >= 1 && @page <= @last or raise(OutOfRangeError, "expected :page in 1..#{@last}; got #{@page.inspect}")
     @offset = @items * (@page - 1) + @outset                              # pagination offset + outset (initial offset)
     @items  = @count % @items if @page == @last                           # adjust items for last page
     @from   = @count == 0 ? 0 : @offset+1 - @outset                       # page begins from item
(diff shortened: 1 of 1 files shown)


lib/pagy.rb
```diff
@@ -28,7 +28,7 @@ def initialize(vars)
       instance_variable_set(:"@#{k}", @vars.delete(k).to_i)               # set instance variables
     end
     @pages  = @last = [(@count.to_f / @items).ceil, 1].max                # cardinal and ordinal meanings
-    (1..@last).cover?(@page) || raise(OutOfRangeError, "expected :page in 1..#{@last}; got #{@page.inspect}")
+    @page >= 1 && @page <= @last or raise(OutOfRangeError, "expected :page in 1..#{@last}; got #{@page.inspect}")
```
MSeneadza: The Ruby style guide frowns upon the use of `or` see https://github.com/bbatsov/ruby-style-guide#no-and-or-or

bquorning: `and` and `or` are perfectly fine for control flow. I would never use them for comparing booleans, e.g. `@page >= 1 and @page <= @last` – *that’s* when bugs start to creep in.

Avdi Grimm explains it much better than I: http://www.virtuouscode.com/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/

ddnexus:
@bquorning: your PR is not just OK, but very welcome! I missed that, and the change not only saves memory but is also faster than `range.cover`.

```ruby
Benchmark.ips do |x|
  x.config(:stats => :bootstrap)

  last = 100000
  page = 100

  x.report('range-cover') do
    (1..last).cover?(page)
  end

  x.report('>= && <=') do
    page >= 1 && page <= last
  end

  x.compare!
end
```

```
Warming up --------------------------------------
         range-cover   309.593k i/100ms
            >= && <=   467.947k i/100ms
Calculating -------------------------------------
         range-cover      5.821M (± 0.7%) i/s -     29.102M in   5.006418s
            >= && <=     13.742M (± 0.4%) i/s -     68.788M in   5.009327s
                   with 95.0% confidence

Comparison:
            >= && <=: 13742355.1 i/s
         range-cover:  5820528.6 i/s - 2.36x  (± 0.02) slower
                   with 95.0% confidence
```

PR #11, opened 2018-05-25
URIs are always joined with forward slash

bquorning:
`File.join` might use backslash on e.g. Windows - URL paths are not File paths.

According to Rack specification:

 - The `SCRIPT_NAME`, if non-empty, must start with `/`
 - The `PATH_INFO`, if non-empty, must start with `/`
 - One of `SCRIPT_NAME` or `PATH_INFO` must be set. `PATH_INFO` should be `/` if
   `SCRIPT_NAME` is empty. `SCRIPT_NAME` never should be `/`, but instead be
   empty.

And in fact Rack provides this handy `#path` method for exactly this purpose: https://github.com/rack/rack/blob/42e48013dd1b6dbd/lib/rack/request.rb#L406-L408

### DIFF
--- a/lib/pagy/frontend.rb
+++ b/lib/pagy/frontend.rb
@@ -34,7 +34,7 @@ class Pagy
 
     # this works with all Rack-based frameworks (Sinatra, Padrino, Rails, ...)
     def pagy_url_for(n)
-      url    = File.join(request.script_name.to_s, request.path_info)
+      url    = request.path
       params = request.GET.merge('page'.freeze => n.to_s)
       url << '?' << Rack::Utils.build_nested_query(pagy_get_params(params))
     end
--- a/pagy.gemspec
+++ b/pagy.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'bundler',  '~> 1.16'
   s.add_development_dependency 'rake',     '~> 10.0'
   s.add_development_dependency 'minitest', '~> 5.0'
+  s.add_development_dependency 'rack'
   s.add_development_dependency 'slim'
   s.add_development_dependency 'haml'
   s.add_development_dependency 'benchmark-ips'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -228,6 +228,7 @@ GEM
     pathutil (0.16.1)
       forwardable-extended (~> 2.6)
     public_suffix (2.0.5)
+    rack (2.0.5)
     rake (10.5.0)
     rb-fsevent (0.10.3)
     rb-inotify (0.9.10)
@@ -273,6 +274,7 @@ DEPENDENCIES
   memory_profiler
   minitest (~> 5.0)
   pagy!
+  rack
   rake (~> 10.0)
   slim
 
--- a/test/frontend_test.rb
+++ b/test/frontend_test.rb
@@ -1,0 +1,72 @@
+require 'test_helper'
+require 'rack'
+
+class FrontendTest < Minitest::Test
+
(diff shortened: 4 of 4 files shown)


lib/pagy/frontend.rb
```diff
@@ -34,7 +34,7 @@ def pagy_info(pagy)
 
     # this works with all Rack-based frameworks (Sinatra, Padrino, Rails, ...)
     def pagy_url_for(n)
-      url    = File.join(request.script_name.to_s, request.path_info)
+      url    = "#{request.script_name}/#{request.path_info}"
```
bquorning: Actually, this is not even needed. According to Rack specification:

> The `SCRIPT_NAME`, if non-empty, must start with `/`
> The `PATH_INFO`, if non-empty, must start with `/`
> One of `SCRIPT_NAME` or `PATH_INFO` must be set. `PATH_INFO` should be `/` if `SCRIPT_NAME` is empty. `SCRIPT_NAME` never should be `/`, but instead be empty.

And in fact `Rack` provides a handy `#path` method for exactly this purpose:

https://github.com/rack/rack/blob/42e48013dd1b6dbda990dfa3851856c199b0b1f9/lib/rack/request.rb#L406-L408

ddnexus: So why don't we use just `url = request.path`?

bquorning: That would be my recommendation as well. I’ll update the PR, perhaps add a few tests.

ddnexus: Thanks

pagy.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'bundler',  '~> 1.16'
   s.add_development_dependency 'rake',     '~> 10.0'
   s.add_development_dependency 'minitest', '~> 5.0'
+  s.add_development_dependency 'rack'
```
bquorning: I don’t know if `rack` should in fact be considered a runtime dependency? You cannot use `Pagy::Frontend` without either having `Rack` loaded, or overriding `Pagy::Frontend#pagy_url_for`.

bquorning:
@ddnexus Ready for review.

pagy.gemspec
ddnexus: > You cannot use Pagy::Frontend without either having Rack loaded, or overriding Pagy::Frontend#pagy_url_for

True, but...

The `pagy_url_for` uses rack just to give a default usable on most apps out of the box. Adding an explicit dependency would cause the installation of rack even if you never intended to use rack in your app and you are fine with just overriding the `pagy_url_for`.

On the other hand not adding it would cause the method to fail until the user will provide his/her own way to create the url, with no feedback added.

The best (puristic) approach would probably be conditionally defining the method in presence of rack, and defining the method raising a `NotImplementedError` exception if Rack is not present.

Something similar to how the `pagy_t` is implemented.

bquorning: Yeah, that sounds like a valid strategy. Except, just use `NoMethodError` instead of `NotImplementedError`: http://chrisstump.online/2016/03/23/stop-abusing-notimplementederror/ (as in: just leave the method not implemented, or raise it with a message about using Rack)

ddnexus: I am fine with both options. Whatever is better for you. Thanks.

bquorning: I agree. I have removed the last commit, so this PR is back at just using `request.path` and adding a few tests where `Rack` needs to be a development dependency.

ddnexus: Thanks

PR #12, opened 2018-05-25
Only use and/or keywords for control flow

bquorning:
Using `and` and `or` for anything other than control flow has too often lead to subtle bugs.


ddnexus:
Good catch! Shame on me! :)

PR #18, opened 2018-05-26
basic rubocop setup

grosser:


ddnexus:
Thanks for the PR. To me it looks OK, but I am not the best reviewer for rubocop, so it would be better if @tibbon @gamafranco @bquorning could chime in.

skatkov:
I don't see anything wrong with rubocop configuration itself. Seems a very good default config (most of questionable stuff is turned off, but really critical things are still being checked).

But I don't think this will be much of a use without integration with CircleCI/Travis to run on every pull request.

gamafranco:
Agree wtih @skatkov:

I would also add bundler-audit and brakemen, failing when a vulnerability is detected. I can do that PR once this one is merged to master.

grosser:
rebased ... merge ?

PR #30, opened 2018-05-30
test on all supported ruby versions

grosser:
picking 2.2+ since everything else is end-of-life ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,13 @@
 language: ruby
 rvm:
+ - 2.3
+ - 2.4
  - 2.5
+script: bundle exec rake $TASK
+env:
+- TASK=test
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,3 +1,7 @@
 language: ruby
 rvm:
+ - 2.2
+ - 2.2
```
ddnexus: duplicate

grosser: updated

ddnexus:
So is this failing because of the missing #31?

grosser:
yeah ... so 2.2 is not supported because it's using `Hash#dig` I'll combine them here for less madness

grosser:
ready!

PR #31, opened 2018-05-30
split rubocop into a separate task so travis failures are easier to p…

grosser:
…arse

will go well with https://github.com/ddnexus/pagy/pull/30

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+AllCops:
+  TargetRubyVersion: 2.2
+
 Layout:
   Enabled: false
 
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -1,3 +1,6 @@
+AllCops:
+  TargetRubyVersion: 2.2
```
grosser: needs to be the minimum version supported or it will recommend things that are not possible

ddnexus: I don't know exactly what is the actual minimum version supported. I developed on 2.5, and I tried to  avoid too modern syntax, but I am not sure I missed something and what the min version would be. Is there any quick-and-easy way to check that?

ddnexus: oh... just saw #30

PR #29, opened 2018-05-30
use actionable code coverage

grosser:
for https://github.com/ddnexus/pagy/issues/13

a rough cut of what it could look like ...

tests will now fail when coverage gaps (also branch coverage) are introduces:

```
# Running:

...
Finished in 0.009591s, 312.7932 runs/s, 312.7932 assertions/s.

3 runs, 3 assertions, 0 failures, 0 errors, 0 skips
lib/pagy/frontend.rb new uncovered lines introduced (10 current vs 0 configured)
Lines missing coverage:
lib/pagy/frontend.rb:20
lib/pagy/frontend.rb:31
lib/pagy/frontend.rb:32
lib/pagy/frontend.rb:33
lib/pagy/frontend.rb:53:64-53:75
lib/pagy/frontend.rb:53:107-53:116
lib/pagy/frontend.rb:69
lib/pagy/frontend.rb:70
lib/pagy/frontend.rb:71
lib/pagy/frontend.rb:72
```

to get this completely working another set of best-practices should be followed:
 - 1 test per lib file (integration tests can be added too)
 - structure test to mirror lib structure
 - do not load all of pagy when loading the gemspec since that would be before coverage reporting can start

... they should be improvements to the discoverability of tests and good coverage on their own, so we can start with them before adding coverage or merge this and then change the structure / improve coverage in parallel

/cc @bquorning @ddnexus @gamafranco

PR #32, opened 2018-05-30
clean up gemspec

grosser:
imo what it should look like, please point out the things that I should revert ...

### DIFF
--- a/pagy.gemspec
+++ b/pagy.gemspec
@@ -1,22 +1,16 @@
-
 lib = File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'pagy'
-require 'date'
 
 Gem::Specification.new do |s|
   s.name          = 'pagy'
   s.version       = Pagy::VERSION
   s.authors       = ['Domizio Demichelis']
   s.email         = ['dd.nexus@gmail.com']
-  s.date          = Date.today.to_s
-
   s.summary       = 'The Ultimate Pagination Ruby Gem'
   s.description   = 'Agnostic pagination in plain ruby: it works with any framework, ORM and DB type, with all kinds of collections, even pre-paginated, scopes, Arrays, JSON data... and just whatever you can count. Easy to use and customize, very fast and very light.'
   s.homepage      = 'https://github.com/ddnexus/pagy'
   s.license       = 'MIT'
-  s.require_paths = ['lib']
-
   s.files         = `git ls-files -z`.split("\x0").select{|f| f.start_with?('lib', 'pagy.gemspec', 'LICENSE') }
   s.required_ruby_version = '>= 2.3.0'
 
@@ -31,11 +25,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'memory_profiler'
   s.add_development_dependency 'rubocop', '~> 0.55.0' # needs to be fixed since it constantly adds new cops
 
-
-  s.post_install_message = <<END_OF_MESSAGE
-******************************************************************************
-   Pagy CHANGELOG: https://github.com/ddnexus/pagy/blob/master/CHANGELOG.md
-******************************************************************************
-END_OF_MESSAGE
-
+  # remove after 1.0
+  s.post_install_message = "Pagy CHANGELOG: https://github.com/ddnexus/pagy/blob/master/CHANGELOG.md"
 end


pagy.gemspec
```diff
@@ -1,25 +1,18 @@
...
-require 'pagy'
-require 'date'
+require 'pagy/version'
 
 Gem::Specification.new do |s|
   s.name          = 'pagy'
   s.version       = Pagy::VERSION
   s.authors       = ['Domizio Demichelis']
   s.email         = ['dd.nexus@gmail.com']
-  s.date          = Date.today.to_s
```
grosser: default

pagy.gemspec
```diff
@@ -1,25 +1,18 @@
...
-
   s.summary       = 'The Ultimate Pagination Ruby Gem'
   s.description   = 'Agnostic pagination in plain ruby: it works with any framework, ORM and DB type, with all kinds of collections, even pre-paginated, scopes, Arrays, JSON data... and just whatever you can count. Easy to use and customize, very fast and very light.'
   s.homepage      = 'https://github.com/ddnexus/pagy'
   s.license       = 'MIT'
-  s.require_paths = ['lib']
```
grosser: default

pagy.gemspec
```diff
@@ -30,12 +23,4 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'kalibera'
   s.add_development_dependency 'memory_profiler'
   s.add_development_dependency 'rubocop', '~> 0.55.0' # needs to be fixed since it constantly adds new cops
-
-
-  s.post_install_message = <<END_OF_MESSAGE
```
grosser: annoying ... always hate these and most other gems thankfully don't do that

pagy.gemspec
```diff
@@ -1,25 +1,18 @@
-
 lib = File.expand_path('../lib', __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require 'pagy'
-require 'date'
+require 'pagy/version'
```
grosser: loading all of pagy is not necessary ... also allows only version to be loaded

pagy.gemspec
ddnexus: The original intention of this message was getting attention on the breaking changes of 0.8.0, and providing a link to the changelog that list them would have been useful. Then the text was too long and I reduced it leaving only the link, and I realize that the original goal is missing now.
A better option would have been something like "Please, check for breaking changes: https://...".
It was meant to be temporary (only for version 0.8.0), but then it would not appear for version after the breaking change and upgrading is not a linear process, so we have to decide whether it is too annoying or somehow useful.
Maybe we should just keep a (better) line of text? Less annoying and still useful?

lib/pagy.rb
```diff
@@ -50,5 +50,6 @@ def series(size=@vars[:size])
 
 end
 
+require 'pagy/version'
```
ddnexus: reuiring the version is not needed here, if you load directly the version in the gemspec

lib/pagy/version.rb
```diff
@@ -0,0 +1,3 @@
+class Pagy
+  VERSION = '0.8.1'
+end
```
ddnexus: I hate having a ruby file for the version

grosser:
signal breaking changes with a major bump :)

pagy.gemspec
ddnexus: agreed

grosser:
for backwards compatability so any user can use Pagy::VERSION
the gemspec gets serialized by rubygems and will no longer include the
require

pagy.gemspec
ddnexus: While true, the "all of pagy" is actually so small that would not bother anyone/anything

pagy.gemspec
ddnexus: agreed

pagy.gemspec
grosser: just a pattern I use/saw a lot and also makes https://github.com/ddnexus/pagy/pull/29 smaller :)

lib/pagy/version.rb
grosser: needed for https://github.com/ddnexus/pagy/pull/29 and kinda a best-practice ...

alternatives:
 - have verison copied in gemspec and lib
 - do not use `gemspec` in Gemfile

ddnexus:
Major bump would be 1.0, which would also indicate API stability to me.... and we are not there yet :)

grosser:
added back a minimal post-install message ... want anything else changed or good to go ?

ddnexus:
do you have 5 mins for a quick [chat on gitter](https://gitter.im/ruby-pagy/Lobby)?

grosser:
nah, on a train ... just post what you want changed :)

ddnexus:
So maybe we can use a `VERSION` file with the version string in it, then pull the string with `File.read('VERSION').strip` in both pagy.rb and gemspec. That is roughly similar to have an extra ruby file, but it will be in the root dir (easier to check in github), and will avoid to have to load the whole code.

ddnexus:
However I still prefer how it is right now: simpler, and really not that bad considering the micro-size of pagy.

grosser:
reverted version handling ... good to go ?

ddnexus:
ooops... I merged #30 first :)

grosser:
np, rebased and ready

PR #33, opened 2018-05-31
avoid uncoverable code by not using gemspec in Gemfile

grosser:
... alternative approach to make code coverage work

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,7 +10,7 @@ Metrics:
 Style:
   Enabled: false
 
-Gemspec/OrderedDependencies:
+Bundler/OrderedGems:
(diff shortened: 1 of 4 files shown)


ddnexus:
Rubocop seems a bit too picky!

ddnexus:
I like moving the dev gems into the Gemfile! Any drawback?

grosser:
fixed rubocop

no drawbacks unless there were any runtime dependencies, then it would duplicate the gemspec ... but I don't think there will ever be runtime dependencies ...

grosser:
merge 32 and I'll rebase it

grosser:
rebased and ready!

PR #37, opened 2018-05-31
improve testability by sorting tests into files that mirror lib + gro…

grosser:
…up by method

makes "where should I test this" straightforward ...

will help with https://github.com/ddnexus/pagy/pull/29


ddnexus:
Nice. Thank you.

PR #40, opened 2018-06-01
improve readability of pagy_link_proc

grosser:
@ddnexus

### DIFF
--- a/lib/pagy/frontend.rb
+++ b/lib/pagy/frontend.rb
@@ -47,13 +47,26 @@ class Pagy
 
     MARKER = "-pagy-#{'pagy'.hash}-".freeze
 
-    # returns a specialized proc to generate the HTML links
-    def pagy_link_proc(pagy, lx=''.freeze)  # "lx" means "link extra"
-      p_prev, p_next, p_lx = pagy.prev, pagy.next, pagy.vars[:link_extra]
-      a, b = %(<a href="#{pagy_url_for(MARKER, pagy)}"#{p_lx ? %( #{p_lx}) : ''.freeze}#{lx.empty? ? lx : %( #{lx})}).split(MARKER)
-      -> (n, text=n, x=''.freeze) { "#{a}#{n}#{b}#{ if    n == p_prev ; ' rel="prev"'.freeze
-                                                    elsif n == p_next ; ' rel="next"'.freeze
-                                                    else                           ''.freeze end }#{x.empty? ? x : %( #{x})}>#{text}</a>" }
+    # returns a performance optimized proc to generate the HTML links
+    def pagy_link_proc(pagy, link_extras=nil)  # "link_extras" means "link extra"
+      p_prev, p_next, page_link_extras = pagy.prev, pagy.next, pagy.vars[:link_extra]
+      page_link_extras = (page_link_extras ? " #{page_link_extras}" : ''.freeze)
+      link_extras = (link_extras ? " #{link_extras}" : ''.freeze)
+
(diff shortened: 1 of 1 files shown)


ddnexus:
Thanks for your PR. While I understand and appreciate your excellent intention, I have quite a few objections to merge it.

- Readability is quite subjective, and I honestly find more difficult to read the proposed method than the original one. You can call me weird :)
- Since it doesn't have to get overridden, and since it is very rarely used by the end-users (i.e. only in case they want to deeply override some `nav_*` method) readability is the least concern
- This method meant to be efficient with performance, memory and code-lines: while the proposed method may subjectively improve the readability (which is not the main goal) I suspect it may slightly reduce its efficiency, but I would have to benchmark each single suspect in order to be sure.

That said, there are a few hints in the proposed method that look better than the original. I will try to  include them. Please, give me some time to digest it :)

grosser:
original looked like a soup to me ... once I de-spaghetied it some patterns were visible and the missing split argument too ... also I could finally understand what was actually going on in there

grosser:
could maybe inline the `#{" #{link_extras}" unless link_extras.empty?}` things since it's still kinda readable ...

grosser:
turns out the `''.freeze` is faster than nil ... which kinda makes sense ...

ddnexus:
I am going to remove all the conditions needed only to remove doble spaces in html tags, which is actually pointless :). After that is should look a lot less like a soup, and actually save one object or two.

grosser:
that sounds good :)

grosser:
also get the `.split(x, 2)`

grosser:
and FYI `if x.empty?` is slower than `if x`

ddnexus:
Yeah, I noticed that. Thanks

ddnexus:
This one should be quite readable:
```ruby
    def pagy_link_proc(pagy, link_extras=nil)
      p_prev, p_next = pagy.prev, pagy.next
      a, b = %(<a href="#{pagy_url_for(MARKER, pagy)}" #{pagy.vars[:link_extras]} #{link_extras}).split(MARKER, 2)
      -> (n, text=n, extras=''.freeze) { "#{a}#{n}#{b}#{ if    n == p_prev ; ' rel="prev"'.freeze
                                                         elsif n == p_next ; ' rel="next"'.freeze
                                                         else                           ''.freeze end } #{extras}>#{text}</a>" }
    end
```

grosser:
Yeah, a lot nicer, but still not a fan of the multiline conditional inside
a string part :)

ddnexus:
I love it! :)
(in that context)

ddnexus:
What about this one:

```ruby
    def pagy_link_proc(pagy, link_extra=''.freeze)
      rel  = Hash.new(''.freeze).merge!(pagy.prev => ' rel="prev"'.freeze, pagy.next => ' rel="next"'.freeze)
      a, b = %(<a href="#{pagy_url_for(MARKER, pagy)}" #{pagy.vars[:link_extra]} #{link_extra}).split(MARKER, 2)
      -> (n, text=n, extra=''.freeze) { "#{a}#{n}#{b}#{rel[n]} #{extra}>#{text}</a>" }
    end
```
It uses one extra hash, but it is as fast and it avoid the multi-line condition in the string. It seems also more readable, and shorter.

grosser:
If it's this or multiline condition in a string then I guess this is better
... but still prefer the `rel = if `

ddnexus:
I went with the first option, because the 2 extra objects would bother me. It is not indented as you would like, but it is a lot more readable now. Thanks or pushing this issue.

PR #48, opened 2018-06-18
Fix typo in :page_param configuration

mcary:
The documentation example for configuring :page_param assigns a symbol
directly to Pagy::VARS.  It should probably assign a symbol to a slot
on Pagy::VARS instead.

Update the documention to assign to "Pagy::VARS[:page_param]" instead.


ddnexus:
Thank you

PR #49, opened 2018-06-19
[Enhance] Default Frozen String for Ruby Files

berniechiu:
### Objective

Since the gem only supports ruby `2.3` and above, I guess it's good enough to start using keyword `frozen_string_literal` to reduce lots of manual freeze methods here

### Feature

1) Replace frozen string methods with keyword `frozen_string_literal`

### DIFF
--- a/lib/pagy.rb
+++ b/lib/pagy.rb
@@ -1,4 +1,5 @@
 # See Pagy API documentation: https://ddnexus.github.io/pagy/api/pagy
+# frozen_string_literal: true
 
 require 'pathname'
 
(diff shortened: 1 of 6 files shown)


ddnexus:
Hi @berniechiu,
there is only a `dig` statement that requires the `2.3`, and since it could be replaced by an `inject`, I was still waiting to decide if I could make it support a few earlier versions. However I should benchmark it to decide whether it is worth the extra versions or not.

I was thinking that the `frozen_string_literal` magic comment should be the very first line in the file. Do it work even after another line?

ddnexus:
It looks like the comment as second line works as well. So well that it breaks the helpers in the bootstrap, compact, items and responsive extras :). They are missing a +

berniechiu:
Thx, I’ll look into it. Cuz I have the test passed locally, weird lol

ddnexus:
Thank you

berniechiu:
sure thx : )

Issue #50, opened 2018-06-20
undefined method `to_i' for Hash in #pagy_get_vars


sliminas:
When having a grouped AR collection the `collection.count(:all)` in [Pagy::Backend#pagy_get_vars](https://github.com/ddnexus/pagy/blob/master/lib/pagy/backend.rb#L20) returns an `OrderedHash` so that `@vars[k].to_i` in [Pagy#initialize](https://github.com/ddnexus/pagy/blob/master/lib/pagy.rb#L22) throws an `undefined method 'to_i' for {1=>1}:Hash` exception.

PR #51, opened 2018-06-20
Add support for counting grouped collections

sliminas:
This PR solves the `undefined method 'to_i' for Hash` exception for grouped AR collections (#50).

### DIFF
--- a/lib/pagy/backend.rb
+++ b/lib/pagy/backend.rb
@@ -16,9 +16,10 @@ class Pagy
 
     # Sub-method called only by #pagy: here for easy customization of variables by overriding
     def pagy_get_vars(collection, vars)
-      # Return the merged variables to initialize the Pagy object
-      { count: collection.count(:all),        # works with AR, but may not work with other ORMs
-        page:  params[vars[:page_param]||VARS[:page_param]] }.merge!(vars)
+      count = collection.count(:all)
+      count = count.count if count.respond_to?(:count) # e.g. AR .group returns an OrderdHash that responds to #count
+      # return the merged variables to initialize the pagy object
+      { count: count, page: params[vars[:page_param] || VARS[:page_param]] }.merge!(vars)
     end
 
     # Sub-method called only by #pagy: here for easy customization of record-extraction by overriding
--- a/test/pagy/backend_test.rb
+++ b/test/pagy/backend_test.rb
@@ -4,7 +4,6 @@ SingleCov.covered!
 
 describe Pagy::Backend do
 
-
   let(:backend) { TestController.new }
 
   describe "#pagy" do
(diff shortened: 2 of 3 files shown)


ddnexus:
Thanks @sliminas. I need to understand it better: couldn't that be avoided by grouping after the pagination? Please, could you post a specific example and use case? Thanks.

KevinColemanInc:
I also just ran into this problem.  It looks like kaminari covers this situation interally.

https://github.com/kaminari/kaminari/blob/4ada38c91128e4a5b9533f00516883d181881d46/kaminari-activerecord/lib/kaminari/activerecord/active_record_relation_methods.rb#L33

KevinColemanInc:
@sliminas 

Can you also update this for the extras as well? they override the `pagy_get_vars` method.

ddnexus:
@KevinColemanInc @sliminas 

To me this looks like quite inefficient: you get it counted WITH distinct so it has to split the total count, and then re-join it, not to mention that you have to check every time whether it is one thing or the other... which doesn't come for free. 

To me counting the full collection with distinct just to have a plain integer looks a bit off.

I asked before and I ask it again because I didn't try it: couldn't that be avoided by groping AFTER the pagination (which would mean that you get a plain count, without splitting and re-joining, and you finally get what you want when you need it?

Thanks

sliminas:
@KevinColemanInc My solution for this was directly taken from how kaminari handles that.
@ddnexus I'll try that and give you an update.

sliminas:
@ddnexus The setup I have:
```ruby
class ProductsController < ApplicationController
  include Pagy::Backend

  def index
    # .... build some other objects depending on the user input
    @products = ProductFilter.filter(dynamic_filter_and_search_variables_and_objects)
     # @products sometimes is a grouped ActiveRecord::Relation
    @pagy, @products = pagy(@products)
  end 

end
```
The grouped collection I get is returned from a `ProductFilter` class which knows how to filter the products according to the search tearm, some filter objects and other parameters. For paginating it before grouping I could not use the very convenient `Pagy::Backend#pagy` mehtod since it's outside of the controller scope. 

So I would have to copy all the logic from `Pagy::Backend` to the `ProductFilter` but I'm not sure if I also want that filter to handle the pagination.

ddnexus:
@sliminas Thanks for the example 
@KevinColemanInc I didn't mean necessarily doing it manually...

BTW, talking about an easy way that doesn't need any PR. It would be enough just defining an alias in the initializer like:

```ruby
class Hash; alias :to_i :count; end
```
That would avoid checking for all the count results. I don't see any practical downside, and in this context it even makes sense: calling `to_i` on a hash would return its count.

KevinColemanInc:
That makes sense.  Maybe just update the docs then?

ddnexus:
The alternatives I tried were not so great. Adding a `to_i` alias to the Hash class was just a bit faster than the proposed PR, but I didn't find any clean way to implement it. 

The original proposed PR has been merged: I just changed the condition using `is_a?(Hash)` instead of `respond_to?(:count)` because it is a bit faster.

@sliminas @KevinColemanInc Thanks for the PR and the feedback.

KevinColemanInc:
@ddnexus 

The PR only fixes it in the default case, not in the add-ins that monkey patch that method.  Could we also get this applied to them as well?

KevinColemanInc:
I meant "extras", I didn't realize that the "parent" method was being called via alias_method:

https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/items.rb#L16

nevermind

PR #58, opened 2018-07-04
Added Turkish locales

msdundar:
Added Turkish locales and tested against:

```ruby
<%== pagy_nav_compact(@pagy) %>
<%== pagy_nav(@pagy) %>
<%== pagy_info(@pagy) %>
```

Since Turkish is an agglutinative language (a way to make new words by adding suffixes to the end of words), it's not possible to make direct translations from English. For example it's not possible directly translate this to Turkish:

`Page 1 of 100.`

For that reason this has translated as `Sayfa 1 / 100.` which sounds and looks perfect in Turkish. Other strings are also tested with zero, one and more item combinations.

### DIFF
--- a/lib/locales/pagy.yml
+++ b/lib/locales/pagy.yml
@@ -22,5 +22,27 @@ en:
     items:
       show: "Show"
       items: "items per page"
+tr:
+  pagy:
+    nav:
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you!

PR #60, opened 2018-07-05
a first small step towards a bmd pagination

wdiechmann:
the pagy_nav_responsive_bmd should build a <ul> but I'm not 'safe' around the js's yet :)

### DIFF
--- a/lib/pagy/extras/responsive.rb
+++ b/lib/pagy/extras/responsive.rb
@@ -69,26 +69,5 @@ class Pagy
       %(<nav id="pagy-nav-#{id}" class="pagy-nav-responsive-bootstrap pagination" role="navigation" aria-label="pager"><ul class="pagination"></ul></nav>#{script})
     end
 
-
-    # Responsive pagination for https://fezvrasta.github.io/bootstrap-material-design: 
-    # it returns the html with the series of links to the pages
-    # rendered by the Pagy.responsive javascript
-    def pagy_nav_responsive_bmd(pagy, id=caller(1,1)[0].hash)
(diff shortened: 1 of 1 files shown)


wdiechmann:
so - I reverted the PR on my fork - hope that will make everything happy-green again :)

ddnexus:
First of all, thanks for your effort!

About the PR, in your case, there should be just one commit that adds the helper, and no merge commit at all, as well as no revert commit at all. I think the best think you could do is just reset the branch and add only the commit that adds the helper... or just start a new PR and add just that.

However what I don't understand about your helper is the use of `span` in place of `ul` + `li` . In the doc that you linked it says

> You can optionally swap out active or disabled anchors for <span>, or omit the anchor in the case of the prev/next arrows, to remove click functionality and prevent keyboard focus while retaining intended styles.

And it looks like the original bootstrap structure of `ul` + `li` is always maintained, while in your helper it has ben replaced by a flat structure. Why?

wdiechmann:
I don't follow - sorry!

Your pagination works by adding span's - and the helper generates a JSON array of that (if I'm not mistaken) so in order to do the least amount of harm, I only exchange the class name given to the spans - hoping that I will be able to style the span's -

But ultimately (once I understand how to change the way the pagy.js works) it will build a list (ul+li)

Pertaining to the commit - perhaps you should just add the method yourself - I don't seem to be able to do a proper PR anyway

ddnexus:
@wdiechmann sorry for the confusion. Reviewing the problem... it looks like the `pagy_nav_responsive_bootstrap` should work as it is with the bmd IMO. Please, try it and let me know.

ddnexus:
Closing this for lack of activity for a week. Reopen it if needed.

PR #66, opened 2018-07-09
Use correct file path for items README

bobbytables:


ddnexus:
Good catch. The fix has been already merged in `dev` with the #63. Thank you.

PR #67, opened 2018-07-09
Fix screenshot name for compact bulma extra docs

enzinia:
Fix Bulma pagination preview for `compact` extra documentation.


ddnexus:
Thanks

PR #71, opened 2018-07-13
use case

grosser:
@ddnexus


ddnexus:
You have been faster than me with the cleanup! Thanks!

Issue #52, opened 2018-06-20
pagy_nav_bootstrap_responsive and CSP unsafe-inline


Bahanix:
Hello, using `pagy_nav_responsive` or `pagy_nav_bootstrap_responsive`, I got this CSP error because of an inline `<script>` injected in the view:

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' […] 'nonce-[…]'". Either the 'unsafe-inline' keyword, a hash ('sha256-[…]'), or a nonce ('nonce-...') is required to enable inline execution.
```

A way to fix without introducing the `inline-safe` vulnerability would be to replace theses `<script>` (eg. [here](https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/responsive.rb#L49)) by something like:

```html
<script type='application/json' class='pagy-responsive'>
{ id: #{id}, tags: #{tags}, widths: #{responsive[:widths].to_json}, series: #{responsive[:series].to_json} }
</script>
```

So one can process them in the main JS, eg: 

```javascript
window.addEventListener("turbolinks:load", function() {
  Array.from(document.getElementsByClassName("pagy-responsive")).forEach(function(pagination) {
    json = JSON.parse(pagination.innerHTML);
    PagyResponsive(json.id, json.tags, json.widths, json.series);
  });
});
```
Is that something Pagy may want? Do I try a PR?

Thanks!

ddnexus:
Hi @Bahanix. Yeah, the inline script maybe a little brutal, I know :)
The reason is there is that is almost self contained, in the sense that the user has only to add the `pagy-responsive.js` and the helper will take are of the rest.

Of course your PR would be welcome! However, please, keep in mind this 2 concepts:

1. the changes should not require any more writing from the user if possible (and I think it is)
2. the changes should be generic as much as possible (e.g. `"turbolink:load"` would be very specific)

If you have different possible alternative about how to implement it, please, feel free to discuss it in the [chat](https://gitter.im/ruby-pagy/Lobby) in advance.

Thanks!

ddnexus:
What about using a `nonce`? Maybe passing it as an extra variable, so it will become totally optional and its generation will be up to the user.

ddnexus:
Actually, the only advantage of a `nonce` is that it wouldn't break the current implementation, but using it is probably more complex than using your approach of removing the inline script all together. 

[...]

PR #73, opened 2018-07-31
Semantic UI nav helper

mad42:
Helper + Doc


mad42:
Hello @ddnexus, really nice work on this gem !

I've just added a test file to complete this first contribution.

I'll add the responsive and compact part in few days, i didn't look at the js part yet.
Happy to help :)

ddnexus:
Thank you @mad42.
No need to look at the js part actually. You have just to do the same as you did for the `pagy_nav` but with the `pagy_nav_compact` and `pagy_nav_responsive`, i.e. mostly changing the classes.

For the `compact` you may probably have to change it a bit more, since it is actually a quite custom helper that you have to render in different ways when using different frameworks. For example, with bootstrap I used a few grouped buttons. That could probably be the same with Semantic UI (which has a lot of nice button components).

mad42:
Helpers for `pagy_nav_compact ` and `pagy_nav_responsive ` embed javascript code, i'm using a very restrictive Content Security Policy forbidding inline javascript `:unsafe_inline` so i need to  isolate the javascript part and hook it differently, it will mess with the existing helpers.

What do you think of the approach of taking off js from helpers ?

Cordially,

ddnexus:
All the helpers pass the :unsafe_inline check since the issue #52 has been closed.

mad42:
Good to know that it was handled before :)
(my fault, i didn't look at previous issues)

I'll try to adapt both helpers !

ddnexus:
Thank you @mad42. Looking forward to merge your final PR.

ddnexus:
Hi @mad42,
what is the status of this PR? I may probably have time for it this weekend...

ddnexus:
Commit cherry-picked, added compact and responsive navs and merged.

Issue #74, opened 2018-08-03
Compatibility with Searchkick?


archonic:
Pagy expects an ActiveRecord scope, but Searchkick returns a Searchkick::Results object. That object responds to most Array methods, or it can return an Array if you call `results` on it.

When using something like this:
```
@pagy, @issues = pagy(
  Issue.search(search_terms, {
    where: {
      open_status: @open_status
    },
    fields: ["title^5", "body^1"],
    select: ["title", "body", "user_id", "created_at", "open_status", "assignee_id"],
    limit: 25,
    load: true
  }).results
)
```
It will complain about no method `offset` and `limit` for Array. I've filled in the blanks with this:
```
class Array
  def offset(value)
    self[value..-1]
  end

  def limit(value)
    self[0..value-1]
  end
end
```
That gets the page to render but it seems there's other missing pieces. All the results are on one page and the navigation renders as plain text.

Has anyone got pagy working with Searchkick? I'm sure lots of folks out there will be trying to mix these 2 great gems.

I noticed the SearchKick::Results object has a bunch of goodies for pagination:
```
 @options=
  {:page=>1,
   :per_page=>25,
   :padding=>0,
   :load=>true,
   :includes=>nil,
   :model_includes=>nil,
   :json=>false,
   :match_suffix=>"analyzed",
   :highlight=>nil,
   :highlighted_fields=>[],
   :misspellings=>false,
   :term=>"*",
   :scope_results=>nil,
   :index_name=>nil},
 @response=
  {"took"=>6,
   "timed_out"=>false,
   "_shards"=>{"total"=>5, "successful"=>5, "skipped"=>0, 
"failed"=>0},
   "hits"=>
    {"total"=>36,
     ...
```

ddnexus:
Hi @archonic,
searchkik results look like a good candidate for an extra, similar to the `array` extra, but specific for searchkik.

archonic:
Working on that now.

PR #75, opened 2018-08-04
Searchkick extra

archonic:
Searchkick has pagination support and that data is available in the `Searchkick::Results` object. I just had to let Pagy know how to read it. It's a whopping 3 lines of difference in method contents from the array extra :raised_hands:.

The tests are a little more terse because Searchkick accepts pagination variables (page and per_page) directly.

### DIFF
--- a/docs/extras/searchkick.md
+++ b/docs/extras/searchkick.md
@@ -1,0 +1,56 @@
+---
+title: Searchkick
+---
+# Searchkick Extra
+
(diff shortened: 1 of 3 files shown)


archonic:
Closes #74

archonic:
@ddnexus Yep, `result.options[:per_page]`. Made that change.

ddnexus:
Thank you!

PR #79, opened 2018-08-07
Foundation extra

dwieringa:
Adds support for Zurb Foundation 6


dwieringa:
Something I'm wondering about now...  I removed the `prev` and `next` classes since I was following the Foundation example (https://foundation.zurb.com/sites/docs/pagination.html) and assumed those were for Bootstrap.  Let me know if those should be included as part of pagy.

Also, last night I took a first pass at adding the compact and responsive versions.  The layout is not right on the compact version and I'm not sure I'll be able to put much more effort in this week.  Do you want me to push this into this PR or wait until I have time to look deeper?

ddnexus:
I see you are making progress! The `prev` and `next` classes do nothing but are there just to add some easy selector in case a user may need it. They are present in all the helpers, so you may want to restore them.
About the `compact` layout... that's the most custom component, hence a bit more tricky to get it right for all the frameworks. Please, push it as it is, if I have some time I may setup a little app and try to help. (if you already have a test app with foundation included, please, let me know so I may save time. Thanks.

dwieringa:
Ok, I added the `prev` and `next` classes back in.  I also committed my `compact` and `responsive` extras even though `compact` needs some layout work.

I don't currently have a simple test app -- I've just been testing inside my client project where I was integrating `pagy`.

By the way, I haven't tested the haml template yet since I don't use it.  I did however compare the output of slim and erb against the helper method.

ddnexus:
Thank you!

ddnexus:
Hi @dwieringa,
I will probably have time to merge this PR during the weekend. What is the status? Just the compact is missing or incomplete?

dwieringa:
Hi @ddnexus,

I'm using the helper in production now and it seems to be working well.
I've compared the erb and slim template output to the helper (but not the haml template).

Correct, in the `compact` extra, the Foundation helper needs work (primarily with layout, which isn't my specialty).  My testing of the `responsive` helper was minimal.

One thought I've had a few times (but I don't have a strong opinion)... would it be better for the compact and responsive helpers to be moved to the respective frontend extra's?  Then the user would choose the frontend extra (eg foundation) and get the default+compact+responsive helpers...vs choosing `compact` and getting default+bootstrap+bulma+foundation+materialize+whatever-comes-later.  This would be a fairly large reorganization though and probably beyond the scope of this PR.

I haven't updated the `compact` and `responsive` docs yet.

ddnexus:
Hi @dwieringa,
thanks for your answer. Yeah, what you are suggesting has been in my mind for a while. There are pros and cons for the 2 types of organizations... however the more new frameworks are coming the more logical would be that change. I will probably go back to this topic as soon as I will find some time.

ddnexus:
Added a few things (ba7d046fb62720f21b0bd120661303762848a008): now it should be complete and work well. Thanks.

ddnexus:
@dwieringa 
FYI: all the extras have been reorganized: the compact and responsive navs have been moved into the respective extras. Thanks

ddnexus:
oh... yes. good catch! Thanks.

PR #82, opened 2018-09-14
Added Spanish locale

EGimenoS:
Added basic spanish translation to the locale pagy.yml.

Tested for pagy_nav, pagy_nav_compact and pagy_info.

For the label "info-single-page-zero" I decided to leave it as "Sin resultados" ("No results"), due to the fact that in spanish words have genre, so if the user is using the name of the resource instead of just "item", incoherent translations might happen ("found" needs to be translated as "encontrados" or "encontradas" depending on the resource name genre.).

On the other hand if the name of the resource is being used instead of "item", the pluralization is not being handled by this translation.

### DIFF
--- a/lib/locales/pagy.yml
+++ b/lib/locales/pagy.yml
@@ -23,6 +23,29 @@ en:
     items:
       show: "Show"
       items: "items per page"
+es:
+  pagy:
+    nav:
+      prev: "&lsaquo;&nbsp;Prev"
+      next: "Siguiente&nbsp;&rsaquo;"
+      gap: "&hellip;"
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you for your PR!
I am surprised about the apparently not translated "item/items". Shouldn't be that translated to "ítem/ítems"?

EGimenoS:
Actually it should, you are correct. I have pushed a new commit in order to fix it.

ddnexus:
Pues, muchísimas gracias!  Es raro, pero de ves en cuando recuerdo poner los acentos donde van! :)

PR #83, opened 2018-09-18
Added color:black style in bootstrap compact input.

Tolchi:
Can't read current page in bootstrap compact navigation due to its color.
Added color: black; in its style.

### DIFF
--- a/lib/pagy/extras/bootstrap.rb
+++ b/lib/pagy/extras/bootstrap.rb
@@ -34,7 +34,7 @@ class Pagy
         html << %(<div class="btn-group" role="group">)
         html << (p_prev ? link.call(p_prev, pagy_t('pagy.nav.prev'), 'aria-label="previous" class="prev btn btn-primary"')
                         : %(<a class="prev btn btn-primary disabled" href="#">#{pagy_t('pagy.nav.prev')}</a>))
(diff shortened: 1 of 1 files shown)


ddnexus:
Hi @Tolchi,
could you post a screenshot?

Tolchi:
before
after
;)

ddnexus:
I don't have the same problem, and since nobody before you raised that issue, I think it is probably due to the CSS rules in your own page OR in the browser you are using. Did you try with other browsers and in a basic app without any CSS customization?

The style that is added to the elements in the helpers and templates tries to be minimal and independent from colors that should belong to the page style, not to the item style.

Is there any reason - that I may be missing - that makes this PR necessary in some context (and would be unharmful to the others)?

serghost:
I have the same issue with bootstrap 3.
screen shot 2018-09-21 at 12 57 08

It's because the input field inherits its color from the btn-primary class (which is #fff) 

Maybe bootstrap 4 doesn't have this issue

ddnexus:
@serghost thank you for the explanation. It makes sense now. We should search a way to avoid the hard-coding of a specific color in the element style. Any idea?

Tolchi:
I´m using bootstrap 3.

ddnexus:
Implemented in a different way (see #86)

Issue #90, opened 2018-09-28
i18n string suggestions


747:
There are a few places where grammatically contiguous variables are outside of translatable string, which makes grammatically correct rendering difficult to impossible in some languages.

For example, this line:
```
 elsif item.is_a?(String) ; %(<li class="current"><span class="show-for-sr">#{pagy_t('pagy.nav.current')}</span> #{item}</li>) # active page
```
is supposed to be displayed in English locale as
> You're on page _#{item}_

while under Japanese word order, it must be like
> 現在 _#{item}_ ページ目です

So, the variable should be included in the string.

-----
Places with same problem as far as I noticed:
```
 html << %(<span class="pagy-compact-input" style="margin: 0 0.6rem;">#{pagy_t('pagy.compact.page')} #{input} #{pagy_t('pagy.compact.of')} #{p_pages}</span> )
```
```
 html << %(#{pagy_t('pagy.items.show')} #{input} #{pagy_t('pagy.items.items')})
```

ddnexus:
True. Very true! :/
Fixing that is probably going to break the current translations, but it is definitely something that needs to be done.
Please, could you post a comprehensive list all of all the places that get affected by that?

PR #93, opened 2018-09-30
I18n refactoring

ddnexus:
As pointed out in #90, a few strings are difficult or impossible to translate in certain languages.

Here are a few commits that re-compose the string in a friendlier way, and remove the `pagy.nav.current` string (marginal and only used by foundation).

Each translation should be reviewed by a native speaker. Particularly the `pagy.items` for `ja`, `ru`, `tr` and `zh-cn`.

Calling in @747, @serghost and @msdundar.


lib/locales/pagy.yml
```diff
@@ -156,12 +141,10 @@ tr:
...
         other: "kayıt"
-    compact:
-      page: "Sayfa"
-      of: "/"
+    compact: "Sayfa %{page_input} / %{pages}"
     items:
-      show: "Göster"
-      items: "sayfa başı"
+      one: "Göster %{items_input} sayfa başı"
+      other: "Göster %{items_input} sayfa başı"
```
msdundar:
```ruby
one: "Sayfada %{items_input} kayıt göster"
other: "Sayfada %{items_input} kayıt göster"
```

is grammatically more correct.

ddnexus: Thank you!

ddnexus: One question: is there any difference if the items in the `items_input` are 1 or more than 1?

msdundar: No, there is not.

lib/locales/pagy.yml
```diff
@@ -178,11 +161,9 @@ zh-cn:
         zero: "项目"
         one: ""
         other: ""
-    compact:
-      page: "页面"
-      of: "/"
+    compact: "页面 %{page_input} / %{pages}"
```
747: Now we can have:
```
    compact: "第 %{page_input} / %{pages} 页"
```

lib/locales/pagy.yml
```diff
@@ -85,19 +76,16 @@ ja:
         zero: "結果"
         one: "結果"
         other: "結果"
-    compact:
-      page: "ページ"
-      of: "/"
+    compact: "ページ %{page_input} / %{pages}"
```
747: So that more I think naturally
```
    compact: "%{page_input} / %{pages} ページ"
```

PR #94, opened 2018-10-09
Update "see extras" broken link

renshuki:
Hi,
I was looking for some info on how to create a Backend Extra and I found this broken link.
Thanks in advance for the merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ _The [IPS/Kb ratio](http://ddnexus.github.io/pagination-comparison/gems.html#eff
 ### Straightforward Code
 
 - Pagy has a very slim core code of just ~100 line of simple ruby, organized in 3 flat modules very easy to understand and use _(see [more...](https://ddnexus.github.io/pagy/api))_
-- It has a quite fat set of optional extras that you can explicitly require for very efficient and modular customization _(see [extras](https://ddnexus.github.io/extras.md))_
+- It has a quite fat set of optional extras that you can explicitly require for very efficient and modular customization _(see [extras](https://ddnexus.github.io/pagy/extras))_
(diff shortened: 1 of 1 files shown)


ddnexus:
Thanks for you PR. What Backend extra are you planning to create?

renshuki:
You are welcome 😄 
Backend extra for elasticsearch-rails or implementing Pagy support for it.
The issue is [here](https://github.com/elastic/elasticsearch-rails/issues/815).

ddnexus:
I read the issue. 

Since Pagy doesn't need to know anything about the collection to paginate, it looks like adding "support" in the collection would not be very useful.

Writing a pagy extra is super simple, especially for the elasticsearch results that contain already the count etc..

I don't know what kind of result is returned by that gem, but if it is a structure like the elasticsearch results I believe that you can take most of the `searchkick` extra to write your own.

renshuki:
Thanks for you feedback.
Indeed it seems way easier to implement this with a pagy extra.
Let's see what @estolfo think about it.

estolfo:
Hi @ddnexus @renshuki Sorry, I'm not following. What is your question exactly?

renshuki:
Hi @estolfo, 
This [issue](https://github.com/elastic/elasticsearch-rails/issues/815) is a request for Pagy support in elasticsearch-rails gem but it's not implemented yet. 

So I was thinking to implement it directly into elasticsearch-rails as you did with will_paginate or kaminari. But then, I dug more into Pagy and I discovered that Pagy have already a really simple way to implement "extras" (consider an extra like a plugin for other gems support). So, I'm thinking to implement elasticsearch-rails support on Pagy side instead.

Just wanted to be sure it's ok for you as well.

Here is the PR #96 if interested.

estolfo:
Hi @renshuki It's certainly ok with me! Thanks for your effort, I'm sure many people would appreciate being able to use Pagy with the `elasticsearch-rails` gem.

Issue #106, opened 2018-11-15
Keyset pagination


espen:
Have you considered [keyset](https://use-the-index-luke.com/no-offset) pagination? It would result in much better database performance.

I see there is [order_query](https://github.com/glebm/order_query) but I haven't tried it. Looking briefly at the code it would definitively complicate things. Perhaps it could be used as a plugin?

What are your thoughts on this?

espen:
I guess what is needed is an attribute for which column to use as a key to order. With an input for last seen value to be used when querying the next page. That sounds simple enough but looking at the order_query source I'm not so sure. I will try it out and see if it can be handled directly in pagy or whether it must be more versatile and relay it to order_query as an extra in pagy.

ddnexus:
Thanks!

ddnexus:
see #136

ddnexus:
Long time since this one.

The [Pagy::Keyset pagination](https://ddnexus.github.io/pagy/docs/api/keyset/) arrived in all its glory in v9.

Yeah... long time... I know 😁

PR #107, opened 2018-11-15
WIP: Pagination without COUNT-sql query

workgena:
### Problem:

Slow pagination for large collection(Database)
"Pagy" make two query:
1. `SELECT * FROM <table> WHERE... LIMIT... OFFSET`
2. `SELECT COUNT(*) FROM <table>`

For large tables (PostgreSQL for example) the `COUNT` works slowly.

### Solution:

Do not show "Total count: X" and do not perform `COUNT(*)` query.

The `extra/infinite` implements this idea.

Usage:

```ruby
  # in initializer of controller
  require 'pagy/extras/infinite'
  require 'pagy/extras/bootstrap' # or other Frontend

  # Controller
  @pagy, paginated_chain = pagy_infinite(Post.all) # `Post` is a model

  # view
  <!-- render Post table -->
  <%== pagy_nav_bootstrap(pagy_var) %>
```


workgena:
This PR might helps improve performance for large tables.
Navigating with large OFFSET is still a performance-issue. But when user has only "next" button, it is hard to reach large offset.

Similar problem discussed in #106

ddnexus:
It looks like a lot of work to come :)
Thanks!

lib/pagy.rb
```diff
@@ -34,6 +34,7 @@ def initialize(vars)
 
   # Return the array of page numbers and :gap items e.g. [1, :gap, 7, 8, "9", 10, 11, :gap, 36]
   def series(size=@vars[:size])
+    return [] if size.empty?
```
workgena: Infinite-extra does not have information does not perform COUNT query. That's why shoring "series" is redundant.
This line make all Frontend render only "Prev | Next" links.

It works for my current project. And it allows easy switching between standard backend and infinite "per-controller".

lib/pagy/extras/infinite.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    # Sub-method called only by #pagy: here for easy customization of variables by overriding
+    def  pagy_infinite_get_vars(collection, vars)
+      current_page = vars[:page] || (params[vars[:page_param] || Pagy::VARS[:page_param]] || 1).to_i
+      per_page_items = vars[:items] || Pagy::VARS[:items]
+      infinite_count = current_page * per_page_items
+      if collection.limit(per_page_items).offset(infinite_count).exists?
+        infinite_count += 1
+      end
+
```
ddnexus: Since the extra is not dedicated to any particular collection type, you should not assume that `limit` and `offset` will be always used for all the collections that might benefit from the extra. You should override the `pagy_get_items` or extracting the collection handling in its own method.

workgena:
The result with `size = []`
image

If the `size` variable set as default [1,4,4,1] then pagination looks like this

| not last page   |      last page      |
|----------|:-------------:|
| image  |   image |

docs/extras/infinite.md
```diff
@@ -0,0 +1,46 @@
...
+
+This extra is composed of 1 file:
+
+- [infinite.rb](https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/infinite.rb)
+
+## Methods
+
+This extra adds the `pagy_infinite` method to the `Pagy::Backend` to be used in place (or in parallel) of the `pagy` method when you have to paginate large table without need of showing "Total count". It also adds a `pagy_infinite_get_variables` sub-method, used for easy customization of variables by overriding.
+
+**Notice**: The `pagy_infinite_get_varuables` returns "count", but it dies not show real COUNT and mustn't be used to show "Total count: N"
```
ddnexus: small typo here

workgena:
Example of a SQL queries made with infinite extra:
````
Contact Exists (0.3ms)  SELECT  1 AS one FROM "contacts" LIMIT $1 OFFSET $2  [["LIMIT", 1], ["OFFSET", 100]]
Contact Load (0.5ms)  SELECT  "contacts".* FROM "contacts" ORDER BY "contacts"."id" DESC LIMIT $1 OFFSET $2  [["LIMIT", 20], ["OFFSET", 80]]
```

And with out it:

```
 (4056.9ms)  SELECT COUNT(*) FROM "contacts"
  ↳ /home/gmalyy/projects/pagy/lib/pagy/backend.rb:19
  Contact Load (0.5ms)  SELECT  "contacts".* FROM "contacts" ORDER BY "contacts"."id" DESC LIMIT $1 OFFSET $2  [["LIMIT", 20], ["OFFSET", 80]]
```

It took 4056.9ms on COUNT(*) query. `contacts` table is 21.1 Million

lib/pagy.rb
workgena: I actually wanted to do it. But `series` methods belongs to `Pagy` itself. And I can't figure out hot wo override it.

lib/pagy/extras/infinite.rb
workgena: I need to think this over. Don't quite understand relation with `pagy_get_items`

lib/pagy.rb
ddnexus: Just define it like:
```rb
class Pagy
  def series(...)
    ...
  end

  module Backend
    ...
  end
end
```

lib/pagy/extras/infinite.rb
ddnexus: Indeed my comment was misleading. I didn't understood your code. :) Forget about it.

So... what you need is checking if there is a full page of items (+1) or not, and that could be done avoiding to use `limit` and `offset` in the method by focusing on the other known variables, which would make the method collection-agnostic.

For example: if you select a page of `:items + 1` and you check that you retrieved `:items + 1` items, then you can pop the last item, so the item number will be correct. And if you pop the last item that means that you can set the `infinite_count += 1`. Instead if the retrieved items is less than `:items + 1` you don't need to pop the last item nor set the `infinite_count`.

Then you should override the `pagy_get_items` and reorganize the `pagy_infinite` accordingly... but I leave it to you for now. ;)

workgena:
Changed

```ruby
if collection.limit(per_page_items).offset(infinite_count).exists?
  infinite_count += 1
end
```

to

```ruby
pagy_next = OpenStruct.new(offset: infinite_count, items: per_page_items + 1)                                                                    
next_collection = pagy_get_items(collection, pagy_next)                                                                                          
if (next_collection.count > per_page_items)                                                                                                      
  infinite_count += 1                                                                                                                            
end
```

The new SQL-queries are:

```
(0.6ms)  SELECT COUNT(*) FROM (SELECT  1 AS one FROM "contacts" ORDER BY "contacts"."id" DESC LIMIT $1 OFFSET $2) subquery_for_count  [["LIMIT", 21], ["OFFSET", 140]]
  ↳ /home/gmalyy/projects/pagy/lib/pagy/extras/infinite.rb:33
  Contact Load (0.5ms)  SELECT  "contacts".* FROM "contacts" ORDER BY "contacts"."id" DESC LIMIT $1 OFFSET $2  [["LIMIT", 20], ["OFFSET", 120]]
```

New query almost as fast as `ActiveRecord#exists?`.

But, to be honest, I don't like `pagy_infinite_get_vars`. It is very complicated. Aspacially

```ruby
current_page = vars[:page] || (params[vars[:page_param] || Pagy::VARS[:page_param]] || 1).to_i                                                   
per_page_items = vars[:items] || Pagy::VARS[:items]
```

I would prefer to read this values from "origina" Backend class.

ddnexus:
@workgena I started to try an alternative solution for this extra. 

It requires one single query, and can use the regular `series` so producing the page-links (for the pages before the current page and even one after it). It can work with regular navs and responsive, but obviously not with the compact and info that require the count.

I don't have much time but I will try to write some documentation, hoping in your help for writing the tests ;). I will push something ASAP.

workgena:
I agree

ddnexus:
See #108

ddnexus:
@workgena if you still have that 20M+ table, you might be interested in the [Pagy::Keyset pagination](https://ddnexus.github.io/pagy/docs/api/keyset/).

Let me know what you think 👀

PR #108, opened 2018-11-18
Countless pagination

ddnexus:
Add the `Pagy::Countless` subclass and the `countless` extra.


ddnexus:
@workgena I didn't try the code yet. If you have time, please, take a look at it. The doc should be already OK, but it is missing the tests for the `Pagy::Countless` and `countless` extra. Any help will be very appreciated. Thanks.

workgena:
Today I'll take a look at this solution. And will write tests for it.

One thing I see strait away is:
 - original `Pagy` returns `ActiveRelation`, in cast of ActiveRecord(Rails)
 - `Contless` returns an `Array`

Since, it is another subclass - it should be OK, no braking change, I think. Worth mention in documentation.

ddnexus:
True! I thought about adding it, and then I forgot. Thanks.

ddnexus:
Thanks for fixing my typos in the items test :).
Thanks for the extra countless tests.
It looks like we miss only the `Pagy::Countless`.

workgena:
Many cases for `Pagy::Countless` already covered(implicitly).

Some new logic in `finalize` should be covered explicitly, values at: 

```ruby
attr_reader :count, :page, :items, :vars, :pages, :last, :offset, :from, :to, :prev, :next
```

workgena:
I've integrated this `countless`-branch into my project at work.

I was a little bit concerned about `COUNTLESS = true`, but it works absolutely fine. Let me explain the use-case:

Each user can decide what pagination to use: Normal(with count) or Fast(Countless) by changing user-settings. So, `PostsController` always has `require 'pagy/extras/countless'` included. Depending on the user settings, the code executes with `pagy(collection)` or `pagy_countless(collection)`.

Both scenarios work as expected.

ddnexus:
I didn't understand what was your concern with `COUNTLESS`.

BTW, your use-case looks interesting. If you have time, you may want to write some article/tutorial explaining what you did.

lib/pagy/extras/items.rb
```diff
@@ -10,14 +10,28 @@ class Pagy
...
+
+    alias_method :pagy_get_vars_without_items, :pagy_get_vars
+    def pagy_get_vars_with_items(collection, vars)
+      pagy_with_items(vars)
       pagy_get_vars_without_items(collection, vars)
     end
     alias_method :pagy_get_vars, :pagy_get_vars_with_items
 
+    # support for countless extra
+    if defined?(Pagy::COUNTLESS)   # defined in the countless extra
```
workgena: Lets imagine PostsController with `require 'pagy/extras/countless'` always.

But in fact it uses normal pagination with `@pagy, @records = pagy(collection)`.

I was afraid, that this block of code populate `vars` with wrong values.

ddnexus:
@espen: just feed `size: []` to remove all the page links, leaving you only the previous/next buttons

workgena:
BTW, me miss this in documentation :books: ...

ddnexus:
@workgena this thime I didn't forget :) [Skipping the page links](https://ddnexus.github.io/pagy/how-to#skipping-the-page-links)

ddnexus:
@espen @workgena Endless scrolling (or incremental scroll) is coming with the `more` extra (currently in development). It will provide the countless pagination with empty series and a few helpers like `pagy_more_url`, `pagy_more_link` and `pagy_more_json` to implement manual or ajax incremental scroll. 

I need someone for trying it and writing tests (hoping to be able to push it soon with at least the doc)

espen:
Thanks. Will try this out a bit more. I think as a user (at least in my app) it's rarely relevant to me how many pages I have. Either I use search or just click next. So skipping one SQL query and not present page links could be a good solution both for performance and UX.

Awesome to hear that endless scrolling will be implemented.

PR #109, opened 2018-11-19
Countless test coverage

workgena:
After reviewing the PR #108 I noticed, that @ddnexus actually wrote a lot of important tests inside `test/pagy/extras/items_test.rb`.

So I cover:
- `pagy_countless_get_vars` method
-  `pagy_countless` with boundary values
-  `vars[:size]` and resulting `series` with filled and empty arrays

If something else is need, I'm ready to help.

PS: @ddnexus You can just Copy/Paste what you need from this PR.

### DIFF
--- a/test/pagy/countless_test.rb
+++ b/test/pagy/countless_test.rb
@@ -16,13 +16,12 @@ describe Pagy::Countless do
 
     let(:last_page) { 3 }
 
-    it 'jump to 2 page with empty collection' do
+    it 'raises exception with no retrieved items' do
       proc { Pagy::Countless.new(page: 2).finalize(0) }.must_raise Pagy::OverflowError
     end
 
-    it 'first page on empty collection' do
+    it 'initializes empty collection' do
       pagy, _ = backend.send(:pagy_countless, @empty_collection, page: 1)
-
       pagy.items.must_equal 20
       pagy.pages.must_equal 1
       pagy.last.must_equal 1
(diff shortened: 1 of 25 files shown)


test/pagy/countless_test.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    before do
+      @empty_collection = TestCollection.new([])
+      @collection = TestCollection.new(Array(1..59))
+    end
+
+    let(:last_page) { 3 }
+
+    it 'jump to 2 page with empty collection' do
+      proc { Pagy::Countless.new(page: 2).finalize(0) }.must_raise Pagy::OverflowError
```
workgena: Here I test actually `finalize()` because `TestCollection` fails with empty array.

test/pagy/countless_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    it 'first page on empty collection' do
+      pagy, _ = backend.send(:pagy_countless, @empty_collection, page: 1)
+
+      pagy.items.must_equal 20
```
workgena: Please take a look on `pagy.items`. It equals 20 on empty collection. And on last page if page is not full, value can be less the 20. For example `pagy.items.must_equal 4` in one test below. Following this logic, for empty collection items should be 0.

ddnexus: That is known and is also documented AFAIK. Quite a long story :).

While the `vars[:items]` remains unchanged, the `@items` is adjusted (for the last non-empty page). The adjustment is needed in order to support the pagination of pre-offsetted and pre-limited collections (let's call them "pO/pL" collections). 

But adjusting the items to 0 for a collection count of 0, would use a limit of 0 in the query, and that would cause a query error with certain DBs. Since adjusting the limit is needed only to retrieve the right items in a pO/pL collection, but when there are 0 items the limit doesn't matter, so it is left unchanged.

workgena: Ok, I read the documentation, but forgot about it.

ddnexus: Now that I think it better... it's probably "documented" only by saying something like "for not-empty pages" or "non-empty collection", but there is not a full explanation of the reason, as in the previous comment. Maybe I should add that somewhere.

ddnexus:
Thank you!

PR #111, opened 2018-11-19
Add polish language support

nowakov:


ddnexus:
Hi @nowakov 
It looks like we should use a different than `:zero_one_other` pluralization rule for `pl`.
Look [here](https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/pl.rb).
It is similar to what we use for `ru` in [:one_few_many_other](https://github.com/ddnexus/pagy/blob/91f2163075d95c59206fd7d1b8de2f58b9ef789f/lib/locales/plurals.rb#L17), but with different numbers.

We should add the `pl` pluralization rule to the `plurals.rb` and reflect it on the `pl.yml`.
Could you do that?

ddnexus:
I just added the support for `pl` in the `plurals.rb` and used it as the base branch for this PR. PLease, update the `pl.yml` accordingly. Thanks.

ddnexus:
Any progress?

nowakov:
@ddnexus I'll try to modify it on weekend

ddnexus:
Added the Polish pluralization to dev. @nowakov please, could you add the missing plurals ('zero', 'one', 'few', 'many', 'other'), so I could finally merge it. Thanks.

ddnexus:
Got the translation in another way. Closing.

PR #112, opened 2018-12-07
if has key "count" in pagy_get_vars, not run double sql query

ermolaev:
```ruby
users = User.social.order(id: :desc)

total_count =
  Rails.cache.fetch("users-count-#{users.query_cache_key}", expires_in: 30.minutes) do
    users.count(:all)
  end

@pagy, @users = pagy(Product.some_scope, count: total_count)
```

### DIFF
--- a/lib/pagy/backend.rb
+++ b/lib/pagy/backend.rb
@@ -16,10 +16,17 @@ class Pagy
 
     # Sub-method called only by #pagy: here for easy customization of variables by overriding
     def pagy_get_vars(collection, vars)
-      count = collection.count(:all)                # work with AR collections: other ORMs may need to change this
-      count = count.count if count.is_a?(Hash)      # fix for the AR grouping count inconsistency (Hash instead of Integer)
+      count =
+        if vars.key?(:count)
+          vars[:count]
+        elsif collection.group_values.any?
+          collection.count(:all).count  # for the AR grouping count inconsistency (Hash instead of Integer)
+        else
+          collection.count(:all)        # work with AR collections: other ORMs may need to change this
+        end
+
       # Return the merged variables to initialize the Pagy object
-      { count: count, page: params[vars[:page_param]||VARS[:page_param]] }.merge!(vars)
+      { count: count, page: params[vars[:page_param] || VARS[:page_param]] }.merge!(vars)
     end
 
     # Sub-method called only by #pagy: here for easy customization of record-extraction by overriding
--- a/test/pagy/backend_test.rb
+++ b/test/pagy/backend_test.rb
@@ -63,6 +63,12 @@ describe Pagy::Backend do
       merged[:link_extra].must_equal 'X'
     end
 
+    it 'get count var' do
+      vars   = {count: 21}
+      merged = backend.send :pagy_get_vars, @collection, vars
+      merged[:count].must_equal 21
+    end
+
     it 'works with grouped collections' do
       @collection = TestGroupedCollection.new((1..1000).to_a)
       vars   = {page: 2, items: 10, link_extra: 'X'}
--- a/test/test_helper/backend.rb
+++ b/test/test_helper/backend.rb
@@ -29,6 +29,10 @@ class TestCollection < Array
     size
   end
 
+  def group_values
+    []
+  end
+
 end
 
 class TestGroupedCollection < TestCollection
@@ -37,4 +41,7 @@ class TestGroupedCollection < TestCollection
     @collection.map { |value| [value, value + 1] }.to_h
   end
 
+  def group_values
+    [:field]
+  end
 end


lib/pagy/backend.rb
```diff
@@ -16,10 +16,17 @@ def pagy(collection, vars={})
 
     # Sub-method called only by #pagy: here for easy customization of variables by overriding
     def pagy_get_vars(collection, vars)
-      count = collection.count(:all)                # work with AR collections: other ORMs may need to change this
-      count = count.count if count.is_a?(Hash)      # fix for the AR grouping count inconsistency (Hash instead of Integer)
+      count =
+        if vars.key?(:count)
+          vars[:count]
+        elsif collection.group_values.any?
+          collection.count(:all).count  # for the AR grouping count inconsistency (Hash instead of Integer)
```
ermolaev: for big performance this code may replaced follow
```ruby
ActiveRecord::Base.connection.select_value(<<~SQL)
  SELECT count(*) FROM (#{collection.except(:select).select('count(*)').to_sql}) t
SQL
```
but requires for test follow dependencies

```
 spec.add_development_dependency 'combustion'
 spec.add_development_dependency 'sqlite3'
 spec.add_development_dependency 'activerecord'
```

ddnexus:
Oh yes! No need to count if there is already a count. How did I miss that?! :)

ddnexus:
Thanks to your suggestion, I implemented it in all the backends (extras included), and extended the behavior to the :page with a little refactoring that saves an extra object and merge.
Thank you!

lib/pagy/backend.rb
ermolaev:
```ruby
require 'benchmark/ips'
scope = Order.where("created_at > '2018-05-05'").group(:user_id)
Benchmark.ips do |x|
  x.report("sub sql") { 
	ActiveRecord::Base.connection.select_value(<<~SQL)
	  SELECT count(*) FROM (#{scope.except(:select).select('count(*)').to_sql}) t
	SQL
  }
  x.report("hash count") { scope.count(:all).count }
  x.report("hash size") { scope.count(:all).size }
  x.compare!
end
```
```
Calculating -------------------------------------
             sub sql      6.291  (± 0.0%) i/s -     32.000  in   5.114788s
          hash count      2.990  (± 0.0%) i/s -     15.000  in   5.028347s
           hash size      2.865  (± 0.0%) i/s -     15.000  in   5.300102s

Comparison:
             sub sql:        6.3 i/s
          hash count:        3.0 i/s - 2.10x  slower
           hash size:        2.9 i/s - 2.20x  slower
```

```ruby
grouped_orders = Order.where("created_at > '2018-05-05'").group(:user_id).count(:all)
Benchmark.ips do |x|
  x.report("hash count") { grouped_orders.count }
  x.report("hash size") { grouped_orders.size }
  x.compare!
end
```

```
Warming up --------------------------------------
          hash count    41.000  i/100ms
           hash size   356.849k i/100ms
Calculating -------------------------------------
          hash count    404.438  (± 9.9%) i/s -      2.009k in   5.021962s
           hash size     11.470M (± 4.4%) i/s -     57.453M in   5.019445s

Comparison:
           hash size: 11469606.0 i/s
          hash count:      404.4 i/s - 28359.35x  slower
```

sub sql 2x times faster, in addition, it uses less memory because it does not create a large hash

ddnexus: I tried it quickly but it couldn't make it work. Is it only for group or for everything? Maybe you could write a small section in the [Dealing with a slow count...](https://ddnexus.github.io/pagy/how-to#dealing-with-a-slow-collection-count), explaining the details...

ermolaev: where and how did you try? could cite some code

sub sql only for queries with grouping, it should work in postgres, mysql, sqlite
ok I will try to write an instruction

also note that the hash.count is very slow, benchmark above

ermolaev:
no, `count(:all)` run sql query and return hash, benchmark tests only methods `size`, `count` without sql overhead, other example

```ruby
grouped_orders = (1..1000).map { |i| [i, i+1] }.to_h
Benchmark.ips do |x|
  x.report("hash count") { grouped_orders.count }
  x.report("hash size") { grouped_orders.size }
  x.compare!
end
```

```
Warming up --------------------------------------
          hash count     1.567k i/100ms
           hash size   298.174k i/100ms
Calculating -------------------------------------
          hash count     16.122k (±15.6%) i/s -     78.350k in   5.001783s
           hash size      9.117M (± 9.4%) i/s -     45.322M in   5.016033s

Comparison:
           hash size:  9117275.2 i/s
          hash count:    16122.0 i/s - 565.52x  slower
```
why big difference?
- method size has complexity o(1)
- method count has complexity o(n)

if grouped_orders has size more 10x `count` will be slow 10x, `size` same speed

lib/pagy/backend.rb
ddnexus: You are totally right! I completely forgot that `count` is meant to iterate over the results of a block and without any block it still counts by iterating. Very expensive! I will change that. Thank you!

ddnexus:
Yes. Changed. Thank you!

PR #116, opened 2019-01-13
Add Dutch language

loed-idzinga:
I was trying out this gem and I noticed that the Dutch language was not present yet, so took the opportunity to add it.


ddnexus:
Thank you!

PR #119, opened 2019-01-31
minor typo corrections

JamesChevalier:
Not much going on here - just some typo fixes that I noticed while reading through the AJAX docs.


ddnexus:
Thank you!

PR #120, opened 2019-02-02
Improve default German translation

thomasklemm:
- Eintrag / Einträge is more idiomatic than Stück / Stücke for records in a list context
- Button labels are capitalized and corrected
- Going to the previous page is labeled "Zurück" now, much more idiomatic in this context

Thanks for this nice gem, @ddnexus :)


ddnexus:
Thank you for your contribution!

PR #125, opened 2019-02-17
Always return the same object as Pagy.root

ashmaroli:
Since the root location is not going to change ever, it can be safely stashed in a constant instead of generating a new `Pathname` object for every call to `Pagy.root`.

Also, `Pathname.new(__dir__)` is the *faster* alternative of `Pathname.new(__FILE__).dirname`. But it won't matter since the computation is going to be **just once** now.


ddnexus:
Yes, a constant would be better! 

At that point, why even bother to have a method returning the constant? I can change that in v2.0, since it is a breaking change.

However, we cannot use `__dir__` in pagy v2, because the ruby version requirement will be lowered to 1.9, and `__dir__` has been added only in ruby v2.0. But, as you said, that doesn't matter.

Thank you for the suggestion!

ashmaroli:
I didn't know development of v2.0 was already underway, so opted for compatibility.
But yes you can do away with `Pagy.root` entirely in v2.0

Should I amend the commit to use `__FILE__`?

ddnexus:
Yes, that would be nice, so it will play better with the replace the `Pagy.root` to `Pagy::ROOT` that I would like to introduce due to your suggestion. Thank you again!

ddnexus:
I have to think better about the constant change... `Pagy.root` actually returns a `Pathname` object (not a simple string) and it would look kind of weird doing `Pagy::ROOT.join(...)`. 

I think that `Pagy.root` is more conventionally used. Rails went from `RAILS_ROOT` (string) to `Rails.root` (Pathname) in version 3 and something and we are all used to that now.

Using a single object would probably reduce a bit the memory usage, even if it is mostly a configuration setting, so your suggestion stands... but I think I will leave the method in place.

ashmaroli:
It doesn't matter either ways as long as you **don't expect** the value returned to be mutable. I've frozen the constant and it is therefore immutable.

ddnexus:
and what about:

```ruby
  def self.root; @root ||= Pathname.new(__FILE__).dirname.freeze end
```

ashmaroli:
> This would even save an object if not used.

:+1:  Lazy initialization's brilliant as well.

PR #127, opened 2019-02-17
Cache gemset for faster Travis builds for v2.0

ashmaroli:
Based on [the documentation](https://docs.travis-ci.com/user/caching/#caches-and-build-matrices), the cache is dependent on the Gemfile's name..


ashmaroli:
Based on the [build log](https://travis-ci.org/ddnexus/pagy/jobs/494479695#L573)s, the caches are indeed separate.
```
uploading cache-linux-trusty-[digest]--rvm-2.0--gemfile-gemfiles/ruby-2.0.tgz
cache uploaded
```

ddnexus:
Brilliant!

ddnexus:
The same change has been merged in `dev` and `v2.0-dev` has been rebased on top of it, so it is already included.

PR #134, opened 2019-03-06
Update example pagy.rb

rafaeelaudibert:
Every extra has its own 'require' line of code, except for the Overflow one, so this PR has the intention to:

Add missing "require 'pagy/extras/overflow'" line of code, to improve beginners usage


rafaeelaudibert:
Excelent job with the gem, btw :smiley:

ddnexus:
Good catch! I cannot believe I missed that for so long :)

rafaeelaudibert:
I must admit it was kinda hidden, as I just noted it when I needed to use it by myself, hahah
I tried to verify all the other 'requires' and they all seem to be there

rafaeelaudibert:
Went through the code and also found a typo on the pt-br.yml file.

ddnexus:
Thanks also for that. I will better wait a bit more before merging so you can catch more glitches ;)

PR #141, opened 2019-03-10
added headers extra

ddnexus:
Add [RFC-8288](https://tools.ietf.org/html/rfc8288) compilant HTTP response headers (and other helpers) useful for API pagination.

- No need for an extra dependency
- No need to learn yet another interface
- It saves quite a lot of memory and CPU
- It works with any pagy object (including `Pagy::Countless`) regardless the type of collection
- It offers more explicit flexibility and simplicity


ddnexus:
@grosser, @bquorning, @gamafranco, @cseelus, @MarieS1415, @NARKOZ, @twnaing... 

You feedback would be appreciated!
Thanks

docs/extras.md
```diff
@@ -5,23 +5,24 @@ title: Extras
...
-| `support`             | Extra support for features like: incremental, infinite, auto-scroll pagination                                                                         | [support.rb](https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/support.rb), [documentation](extras/support.md)                                     |
-| `trim`                | Remove the `page=1` param from links                                                                                                                   | [trim.rb](https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/trim.rb), [documentation](extras/trim.md)                                              |
+| Extra                 | Description                                                                                                                               | Links                                                                                                                                                        |
```
grosser: could avoid the giant diff / make blame simler by not aligning the `|`

docs/extras/headers.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+Or if you can reuse it (e.g. in a pure API app with very standard actions), you are encouraged to define a custom `pagy_render` method in your application controller like:
+
+```ruby
+def pagy_render(collection, vars={})
+  pagy, records = pagy(collection, vars) # any pagy_* backend constructor works
+  pagy_headers_merge(pagy)
+  render json: records
+end
```
grosser: could be more universal to do
```Ruby
def render(*args, &block)
  pagy_headers_merge(@pagy) if @pagy
  super
end
```

docs/extras/headers.md
```diff
@@ -0,0 +1,104 @@
...
+
+# and use it in your standard actions:
+pagy_render(Product.all)
+```
+
+## Files
+
+This extra is composed of 1 small file:
```
grosser: this comment is kinda redundant

lib/pagy/extras/headers.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    include Helpers
+
+    def pagy_headers_merge(pagy)
+      response.headers.merge!(pagy_headers(pagy))
+    end
+
+    def pagy_headers(pagy)
+      hash = pagy_headers_hash(pagy)
+      { 'Links'    => hash[:links].map{|rel, link| %(<#{link}>; rel="#{rel}")}.join(', ') }.tap do |h|
```
grosser: more readable/performant to not use `.tap`

lib/pagy/extras/headers.rb
```diff
@@ -0,0 +1,35 @@
...
+        h['Count'] = h['Total']    = hash[:count] if hash.key?(:count)
+      end
+    end
+
+    def pagy_headers_hash(pagy)
+      countless = defined?(Pagy::Countless) && pagy.is_a?(Pagy::Countless)
+      rels      = { first: 1, prev: pagy.prev, next: pagy.next }.tap{ |r| r[:last] = pagy.last unless countless }
+      a, b      = pagy_url_for(Frontend::MARKER, pagy, :url).split(Frontend::MARKER, 2)
+      links     = Hash[rels.map{|rel, n|[rel, "#{a}#{n}#{b}"] if n}.compact]
+      { links: links }.tap do |h|
```
grosser: same here, prefer no tap

lib/pagy/extras/headers.rb
```diff
@@ -0,0 +1,35 @@
...
+      { 'Links'    => hash[:links].map{|rel, link| %(<#{link}>; rel="#{rel}")}.join(', ') }.tap do |h|
+        h['Items'] = h['Per-Page'] = hash[:items]
+        h['Count'] = h['Total']    = hash[:count] if hash.key?(:count)
+      end
+    end
+
+    def pagy_headers_hash(pagy)
+      countless = defined?(Pagy::Countless) && pagy.is_a?(Pagy::Countless)
+      rels      = { first: 1, prev: pagy.prev, next: pagy.next }.tap{ |r| r[:last] = pagy.last unless countless }
+      a, b      = pagy_url_for(Frontend::MARKER, pagy, :url).split(Frontend::MARKER, 2)
```
grosser: got some more descriptive names than a/b ?

lib/pagy/extras/headers.rb
```diff
@@ -0,0 +1,35 @@
...
+      links     = Hash[rels.map{|rel, n|[rel, "#{a}#{n}#{b}"] if n}.compact]
```
grosser: if none of them is nil this will be nil ?

```
[].compact!
=> nil
```

lib/pagy/frontend.rb
```diff
@@ -10,10 +10,23 @@ class Pagy
   # see https://ddnexus.github.io/pagy/api/frontend#i18n
   I18n = eval(Pagy.root.join('locales', 'utils', 'i18n.rb').read) #rubocop:disable Security/Eval
 
+  module Helpers
+    # This works with all Rack-based frameworks (Sinatra, Padrino, Rails, ...)
+    def pagy_url_for(page, pagy, path_or_url=:path)
```
grosser:
```suggestion
    def pagy_url_for(page, pagy, path_or_url: :path)
```

lib/pagy/frontend.rb
```diff
@@ -10,10 +10,23 @@ class Pagy
...
+      p_vars = pagy.vars; params = request.GET.merge(p_vars[:page_param].to_s => page).merge!(p_vars[:params])
```
grosser: can avoid a merge:
```suggestion
      p_vars = pagy.vars; params = request.GET.merge(p_vars[:params])
      params[p_vars[:page_param].to_s] = page
```

test/pagy/extras/headers_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    it 'returns the full headers hash' do
+      pagy, _records = @controller.send(:pagy, @collection)
+      @controller.send(:pagy_headers_merge, pagy)
+      @controller.send(:response).headers.must_equal({"Links"=>"<https://example.com:8080/foo?page=1>; rel=\"first\", <https://example.com:8080/foo?page=2>; rel=\"prev\", <https://example.com:8080/foo?page=4>; rel=\"next\", <https://example.com:8080/foo?page=50>; rel=\"last\"", "Items"=>20, "Per-Page"=>20, "Count"=>1000, "Total"=>1000})
+    end
+
+  end
+
+
+
```
grosser: extra newlines ?

docs/extras.md
ddnexus: true, but a non-aligned source makes me sick :)

docs/extras/headers.md
ddnexus: true! This one is a nice minimalist way to just add the headers automatically that works also with existing code. The other one is more a suggestion about encapsulating/decluttering. I will use both suggestions. Thanks

docs/extras/headers.md
ddnexus: true. I will remove the stupid comments from all the extras and leave only the links to the files. Thanks

lib/pagy/extras/headers.rb
ddnexus: probably... I will try

lib/pagy/extras/headers.rb
ddnexus: not sure to understand "If none of them": did you mean "if all them are nil"?
BTW, there will be always at least the first page link (even if it is empty), so that should not happen... however I will try to avoid asumptions. Thanks

lib/pagy/frontend.rb
ddnexus: nicer, but Pagy is ruby 1.9+ compatible since version 2 (I needed to use it in legacy apps with legacy envs)

lib/pagy/frontend.rb
ddnexus: we didn't notice that for so long! Thank

test/pagy/extras/headers_test.rb
ddnexus: ooops

lib/pagy/extras/headers.rb
ddnexus: like `part_a` and `part_b`? :smile:

grosser: was hoping they have a meaning like url/query "MARKER" is also vague :D

lib/pagy/extras/headers.rb
grosser: if no `nil` is removed the result will be nil

lib/pagy/extras/headers.rb
ddnexus: well, MARKER means... it's just a MARKER :smile: ... used to split the string into 2 parts and wrap them around the page number. That was faster than using `sub` to replace MARKER with the page number (at least in `pagy_link_proc`), but the 2 parts are just 2 strings without any logical meaning. It's difficult to give them a meaningful name.

lib/pagy/extras/headers.rb
ddnexus: I still don't get your comment at all :/ (maybe my morning-dumbness... or just dumbness)
How `[].compact!` relates to the code there? Is that a suggestion or a warning?

espen:
Didn't see this until now. Works great. Thanks 👏

PR #142, opened 2019-03-10
Use "Link" instead of "Links" for header extra

cseelus:
Just looked at the `README` for the headers feature and it really looked simple enough, so I couldn’t resist testing it.

Purging the api-pagination gem and replacing the API stuff was really straight forward, simple and quick.

At first glance with Postman everything looked ok, so I started up an App in iOS Simulator that consumes the API and paging didn’t work. 
So looking closer at the links Pagy injects, I saw that you use 
```
    Links →<…
```
instead of
```
    Link →<…
```

I think it would be hard to justify changing Apps that consume the API(s) the headers extra could be used for and [RFC 8288](https://tools.ietf.org/html/rfc8288) (page 14) also uses Link like api-pagination, so you might reconsider this.

### DIFF
--- a/docs/extras/headers.md
+++ b/docs/extras/headers.md
@@ -51,14 +51,14 @@ This extra is composed of 1 small file:
 
 ## Headers
 
-This extra adds the standard/legacy headers used by various API-pagination gems with `Links`, `Per-Page` and `Total` headers, so it is a convenient replacement for legacy apps that use external gems.
+This extra adds the standard/legacy headers used by various API-pagination gems with `Link`, `Per-Page` and `Total` headers, so it is a convenient replacement for legacy apps that use external gems.
 
 For new apps, and for consistency with the Pagy naming, you may want to use the `Items` (instead of Per-Page) and `Count` (instead of Total) aliases.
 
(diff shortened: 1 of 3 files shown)


ddnexus:
ooops... my bad. For some reason I had a sort of mind-typo that made me read and write Links in place of Link everywhere!

BTW I am not so sure about the naming of the other headers in #141. Is there any standard or should them be considered custom headers? Some recent gem use 'X-Total-Page' and 'X-Total-Count' and for custom headers the 'X-' prefix looks more correct to me. 

I think I am going to add a `:headers_map` var or something to give the user the possibility to custom-name the headers how they want, but the default naming still puzzles me. Any suggestion?

cseelus:
Checked the new head of feature/headers and paging now works as expected.

There is nothing in the RFC regarding this. From what I've seen api-pagination gem uses `Total` while api-pager uses `X-Total-Count`.

The latter (prefixing `X-` kind of makes more sense, but I saw this was [deprecated a while ago](https://tools.ietf.org/html/rfc6648).

ddnexus:
I was not aware of the deprecation link. Thank you!

PR #154, opened 2019-04-24
add french translation

MrMoins:
Let's add french translation

### DIFF
--- a/lib/locales/fr.yml
+++ b/lib/locales/fr.yml
@@ -1,0 +1,20 @@
+fr:
+  pagy:
+    nav:
+      prev: "&lsaquo;&nbsp;Précédent"
+      next: "Suivant&nbsp;&rsaquo;"
+      gap: "&hellip;"
+    info:
+      single_page:
+        zero: "Aucun %{item_name} trouvé"
+        one: "<b>1</b> %{item_name}"
+        other: "<b>tous les %{count}</b> %{item_name}"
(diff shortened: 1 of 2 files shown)


lib/locales/fr.yml
```diff
@@ -0,0 +1,20 @@
+fr:
+  pagy:
+    nav:
+      prev: "&lsaquo;&nbsp;Précédent"
+      next: "Suivant&nbsp;&rsaquo;"
+      gap: "&hellip;"
+    info:
+      single_page:
+        zero: "Aucun %{item_name} trouvé"
+        one: "<b>1</b> %{item_name}"
```
ddnexus: In the French language pluralization the `one` entry is used from 0 up to 2, as I see [here](https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/one_upto_two_other.rb). 

Since the pagy dictionaries should be I18n compatible, shouldn't we have only the `one` entry that should work for 0, 1 and 2?

Here are 2 suggestions (I have no idea which one would work for 0.1 and 2)
```suggestion
        one: "<b>%{count}</b> %{item_name}"
```

```suggestion
        one: "<b>%{count}</b> %{item_name} trouvé"
```

ddnexus: I think that I was wrong... indeed 2 should return 'other' so no need to change anything. 
I thought 2 was included into 'one' but it is not. D'oh :)

ddnexus:
Thank you!

PR #155, opened 2019-04-26
Add french localization file

czj:
We would love to use pagy in production on our (French) apps.

We hope this localization will help other teams use this great gem ❤️

### DIFF
--- a/lib/locales/fr.yml
+++ b/lib/locales/fr.yml
@@ -1,0 +1,20 @@
+fr:
+  pagy:
+    nav:
+      prev: "&lsaquo;&nbsp;Précédent"
+      next: "Suivant&nbsp;&rsaquo;"
+      gap: "&hellip;"
+    info:
+      single_page:
+        zero: "Aucun %{item_name} trouvé"
+        one: "Affichage de <b>1</b> %{item_name}"
+        other: "Affichage de <b>tous les %{count}</b> %{item_name}"
+      multiple_pages: "Affichage des %{item_name} <b>%{from} à %{to}</b> sur <b>%{count}</b> au total"
+      item_name:
+        zero: "élements"
+        one: "élement"
+        other: "élements"
+    compact: "Page %{page_input} sur %{pages}"
+    items:
+      one: "Afficher %{items_input} élément par page"
+      other: "Afficher %{items_input} élements par page"


ddnexus:
Thank you for the contribution!
I just merged another PR for French a couple of days ago. Here is the commit https://github.com/ddnexus/pagy/pull/154/commits/cf425d8b81056b1088fbd57871602d5e21c82de1.
I notice that they use different terms, and my French is not good enough to understand which one would be better.

Could you give me your feedback? Thanks!

ddnexus:
It looks like your translation is closer to the original en text.

czj:
Sorry, I did not see the PR.

Our PR is indeed written in a less "natural" french langage.
We had less experience with Pagy usage/location of messages.

Thanks to https://github.com/MrMoins for the translations !

ddnexus:
Wait. Are you saying that yours is less natural? I actually thought otherwise. To me it looked like it was more consistent with Pagy. Now I am confused. :/

czj:
You are right, in both sentences :-)

- my translation is closer to Pagy’s « technical » terms, because you use the word « item » which translates as « élément »
- his translation is closer to usual terms in french langage : there’s often « search results » that translates to « résultats de recherche »

Depends on what you prefer using !

ddnexus:
Thanks for the explanation... 

then since a list may also be something not related to a search, I think that a more generic élément could be more consistent with both a result from a search or a simple listing (so résultat seems more weird unless you did search). What do you think?

What about "Affichage" and "Afficher" that are missing in the other translation? That is what made me think that is more close to convey the same meaning than in English, but I would like to know what you think about that too.

czj:
> then since a list may also be something not related to a search, I think that a more generic élément could be more consistent with both a result from a search or a simple listing (so résultat seems more weird unless you did search). What do you think?

I agree, a more generic term works better (that's why you had it in English I presume)

> What about "Affichage" and "Afficher" that are missing in the other translation? That is what made me think that is more close to convey the same meaning than in English, but I would like to know what you think about that too.

"Affichage" can be translated a "the display of" more than "displaying" (that's a French specific way of "stating".

"Afficher" is the verb. "Afficher 50 éléments" would be a short / imperative version of "Je voudrais afficher 50 éléments". (which is never used). It's giving the order to "display 50 items".

I see you have used my translation in 3.0 :-) Thanks.

ddnexus:
> I see you have used my translation in 3.0 :-) Thanks.

Thank you for the PR and for the explanations. The other translation is OK too, just a bit too specific to be used as the generic default IMO. Anyway, everybody can easily change it as it prefers, so it is not a concern.

Issue #110, opened 2018-11-19
UIkit template


nowakov:
Hey!
Great gem, but I'd love to have UIkit support built in :) (https://getuikit.com/docs/pagination)
Will you consider merging it if I provide a PR?

Cheers,
Matt

ddnexus:
Your PR would be very welcome!
You can start from another frontend extra.
Thanks

ddnexus:
Closing this for inactivity

PR #164, opened 2019-06-19
Implement UIKit pagination

lfidnl:
Hello!
I've implemented uikit nav styles. It can be linked with https://github.com/ddnexus/pagy/issues/110


ddnexus:
Thank you! Could you also add a documentation page and link it in the other doc and README pages that mention the frontend extras? You could use the `bootstrap` extra as the base for the doc, and search where it is used for references and add the uikit extra too. Thank you.

lfidnl:
@ddnexus OK! I will add docs

lfidnl:
@ddnexus I've completed documentation. Please, let me know if I missed something

ddnexus:
You are fast! 
There are probably just a few references to the templates and in the repo README if I recall correctly. See what you can find, but don't worry about missing anything, I will double check before merging it. Thank you!

lfidnl:
I've added missed reference in README. I hope these were last pushes ;-)

ddnexus:
Thank you @lfidnl. I will review it and merge it during the weekend.

ddnexus:
I added one commit, and I noticed only now that there is no `pagy_uikit_nav_js` and `pagy_uikit_combo_nav_js` helpers. 

Please, let me know if you can add them, because they are quite faster than the regular `pagy_nav` and all the other frontend extras supports them.

They are not difficult to add... the only potentially difficult part would be styling the combo_nav consistently with the uikit (which I should install in some app). Please, let me know.

lfidnl:
Ok! I'll try to add *_js helpers

lfidnl:
@ddnexus hello. I've added *_js methods. Please, check it. I hope I make it right

ddnexus:
Apparently the jruby RVMs have some problem on TravisCI. That's not the first time. Let's wait a bit...

ddnexus:
Thank you @lfidnl! It looks good!

I noticed that the gif for the combo nav is a bit weird: there is a grey border only on the top, and the background is white instead of `#F7F7F7` (as for the docs background). 

Screenshot from 2019-07-09 19-09-22

Please, let me know whether that intended or it is just a glitch of the screenshot. Thanks.

lfidnl:
Sorry, that was my fault. Now I fixed screenshot (added gap in 3px at all sides)

PR #169, opened 2019-07-21
Add term default value for case when term not passed

bestwebua:
Add the same behaviour like in ```Searchkick.searchkick_search``` for ```term``` default value.

### DIFF
--- a/lib/pagy/extras/searchkick.rb
+++ b/lib/pagy/extras/searchkick.rb
@@ -22,6 +22,7 @@ class Pagy
 
     # Return Pagy object and results
     def pagy_searchkick(search_args, vars={})
+      search_args.insert(1, '*') && search_args.delete_at(-2) unless search_args[1].is_a?(String)
       model, term, options, block, *called = search_args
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you for the PR!

I refactored it in a different way that has a few advantages compared to adding a forced `'*'` as a term.

The refactoring is more agnostic of the way searchkick (and elasticsearch_rails) work. The only assumption is that there must be a last argument that must be a hash (which is what pagy use as the options), so it would modify less the original passed arguments. 

For example: if you pass a single hash it will still pass a single hash to the search method (adding the pagination options to it), so it should be a bit more robust in case of upstream changes.

I also added a few tests to cover the optional term argument.

Please, can you try it (dev branch). Thanks.

bestwebua:
Hi, Domizio! Works like a charm! I have checked with tems parameter and without it. As expected, the output results were the same :+1:

ddnexus:
@bestwebua thank you for your help!

Pushed to rubygem as version 3.4.1.

bestwebua:
Thank you, bro!

PR #172, opened 2019-08-04
Add translation ko.yml for korean

yhk1038:
Thanks to make the awesome gem. 👍(duplicate commit inside a last one)

### DIFF
--- a/lib/locales/ko.yml
+++ b/lib/locales/ko.yml
@@ -1,11 +1,9 @@
-# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+# :other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
 
 ko:
   pagy:
 
-    item_name:
-      one: "항목"
-      other: "항목들"
+    item_name: "항목"
 
     nav:
       prev: "&lsaquo;&nbsp;이전"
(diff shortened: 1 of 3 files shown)


lib/locales/ko.yml
```diff
@@ -0,0 +1,22 @@
+# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+
+ko:
+  pagy:
+
+    item_name:
+      one: "항목"
```
ddnexus: I think `:one` is not used in the `ko` pluralization (which should be `:other` type)

lib/locales/ko.yml
```diff
@@ -0,0 +1,22 @@
+# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
```
ddnexus: I think this pluralization should be `:other` and not `:one_other`

ddnexus:
Hi @yhk1038, thank you for the PR.

According to the `ko` rails pluralization [file](https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/ko.rb) the pluralization to use should be `:other`.

lib/locales/ko.yml
yhk1038: Oh, i just skipped this line means. I'll check my fault with the other comment below and PR again. Thank your review 👍

ddnexus:
It looks good now. Thanks for your contribution!

PR #176, opened 2019-08-16
update how-to documentation

BKSpurgeon:
**Why this change?**

Because people may cut and paste (as I did) without remembering to limit the gem to a particular version - to avoid the situations where there are breaking api changes which may catch people unawares. So I thought it best to add a version limitation.

I hope this helps.


ddnexus:
Thank you for the suggestion. Implemented slightly differently in `dev`.

PR #177, opened 2019-08-16
add example

BKSpurgeon:
To make it explicitly clear to newbies (like me) how to quickly get started with pagy - with a concrete example of what and where to paste things. The example will hopefully still be clear enough to other developers who are using frameworks other than rails.

I hope this helps.


ddnexus:
Thank you for the suggestion. Implemented slightly differently in `dev`.

PR #181, opened 2019-08-25
refactor:      extract code into a method to enhance readability

BKSpurgeon:
**Why this change?**

Perhaps this might improve readability, to make it easier for those unfamiliar to the library to read and understand it, and to contribute/modify the code if required. I can contribute more PRs if this change is welcome. 

In any case, I hope this PR is helpful. If not, feel free to scrap it.

regards

Ben

### DIFF
--- a/lib/pagy.rb
+++ b/lib/pagy.rb
@@ -16,11 +16,11 @@ class Pagy ; VERSION = '3.5.0'
 
   # Merge and validate the options, do some simple arithmetic and set the instance variables
   def initialize(vars)
-    @vars = VARS.merge(vars.delete_if{|_,v| v.nil? || v == '' })               # default vars + cleaned vars
+    @vars = clean_vars(vars)
     { count:0, items:1, outset:0, page:1 }.each do |k,min|                     # validate instance variables
       (@vars[k] && instance_variable_set(:"@#{k}", @vars[k].to_i) >= min) \
-         or raise(VariableError.new(self), "expected :#{k} >= #{min}; got #{@vars[k].inspect}")
-    end
+        or raise(VariableError.new(self), "expected :#{k} >= #{min}; got #{@vars[k].inspect}")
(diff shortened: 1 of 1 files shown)


ddnexus:
Ben, thanks for your idea, however there are a couple of drawbacks:

1. pagy will have to call a one-line method from one single place every time an object is created: calling a method instead of just executing its code adds a (minimal) cost, nevertheless there is no need to pay it
2. I am not sure that adding a method name vs using an explanatory comment like `# default vars + cleaned vars` is actually improving code understanding, and in any case the priority is still performance

BKSpurgeon:
ok chrs for reviewing. I appreciate.

> I am not sure that adding a method name vs using an explanatory comment like # default vars + cleaned vars is actually improving code understanding,

Yes the readability in this case is minor.  But my real intention was to improve the readability of this bit (see below image): it's a little intimidating for those unfamiliar with the code base. But there will be small performance overheads. Shall I attempt it?

image

ddnexus:
Admittedly, that are a 3 dense lines... and I just realized that the comment is missing "defines and..." at the beginning, so I will add that.

However that is a very efficient way to set and validate the variables. It keeps the number of objects created (and the lines of code) as low as possible. Of course you need a bit more focus to read it, but it's pure ruby. It's not calling a bunch of methods that you know nothing about: you can understand everything it does right in the 3 lines without knowing nothing else about pagy.

Besides it's not an example in a tutorial where the goal is explaining some ruby code. Here the goal is to be fast and lite. If you are trying to subclass it and you have to go through its details, then you can refactor it in your own class (at least temporarily, in order to study it "your way"). Or if you think that explaining the code piece by piece would be useful to somebody, you could write some documentation about it, but frankly I would consider to change that only in the presence of some benchmarks showing that it would be faster, or lighter, or more readable with the same performance.

PR #185, opened 2019-08-29
Adding elasticsearch7 support

### LINKED ISSUE #184: Elasticsearch 7 support
I'm using elasticsearch 7, and it looks like the `total` attribute in the response is [now a hash](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-request-body.html#request-body-search-track-total-hits) instead of an integer count. So we'd probably need to extract the count using something like:
```
  total = response.raw_response['hits']['total']
  total.is_a?(Hash) ? total['value'] : total
```

I'm happy to open up a PR, but wanted to know your thoughts on this, and if we should add an additional mock model for ES7.

amenon:
Elasticsearch 7 changed the way it returns the [`total` attribute in the response](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-request-body.html#request-body-search-track-total-hits). It is now a hash instead of an integer. This PR adds support for the hash.

Fixes #184

### DIFF
--- a/lib/pagy/extras/elasticsearch_rails.rb
+++ b/lib/pagy/extras/elasticsearch_rails.rb
@@ -13,7 +13,8 @@ class Pagy
   def self.new_from_elasticsearch_rails(response, vars={})
     vars[:items] = response.search.options[:size] || 10
     vars[:page]  = (response.search.options[:from] || 0) / vars[:items] + 1
-    vars[:count] = response.raw_response['hits']['total']
+    total = response.raw_response['hits']['total']
+    vars[:count] = total.is_a?(Hash) ? total['value'] : total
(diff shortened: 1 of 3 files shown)


ddnexus:
Thank you!!!

exocode:
sorry for reposting this here on the PR 
Is it possible to add this feature also to version 3.x, because Ruby 3.x is quite "experimental" (I had issues to debug rspec) installing nokogiri and so on.. so it is - in my opinion - not ready for production.

So it would be great to make this feature available to Ruby 2.x users too...

ddnexus:
This feature should already be in 3.x since almost 2 years. Do I miss something?

PR #188, opened 2019-09-09
Update how-to: Correct to require headers for API

yisselda:
I was going through the doc, if you don't include headers, you won't get the method `pagy_headers_merge` since it's not in metadata.


ddnexus:
Right! Thank you!

PR #190, opened 2019-09-15
remove order from scope

pedrocarmona:
When the select statement has a alias for a table/formula, and we order by it, then rails `count(:all)` will still use the order. In my case this makes it fail. Removing the order makes it work.

### DIFF
--- a/lib/pagy/backend.rb
+++ b/lib/pagy/backend.rb
@@ -18,7 +18,7 @@ class Pagy
 
     # Sub-method called only by #pagy: here for easy customization of variables by overriding
     def pagy_get_vars(collection, vars)
-      vars[:count] ||= (c = collection.count(:all)).is_a?(Hash) ? c.size : c
+      vars[:count] ||= (c = collection.unscope(:order).count(:all)).is_a?(Hash) ? c.size : c
       vars[:page]  ||= params[ vars[:page_param] || VARS[:page_param] ]
       vars
     end
--- a/test/mock_helpers/collection.rb
+++ b/test/mock_helpers/collection.rb
@@ -18,6 +18,10 @@ class MockCollection < Array
     size
   end
 
+  def unscope(*)
+    self
+  end
+
   class Grouped < MockCollection
 
     def count(*)


ddnexus:
This is interesting. Thank you.
 
However `unscope`  applies only to a small range of rails version according to [ this](https://apidock.com/rails/v3.1.0/ActiveRecord/Base/unscoped/class). Besides, not all collections are ordered, so that would apply only to a subset of apps.

It looks like merging that change in the standard code-base would break a lot of apps.

I believe that in your case, the canonical way to avoid the error would be passing the custom-count to the pagy constructor.

I never thought about that before, but it looks like AR/arel is not smart enough to remove the useless scopes (like order) when it has to count a collection. I don't have time to try that now, but if it does not remove the order scope by itself when a count is requested, then it could be a nice improvement to suggest for Arel.

pedrocarmona:
Hi Domizio, thanks for feedback. I totally agree. Both solutions are great, thank you.

lostapathy:
@ddnexus any chance you'd consider a patch like this in 2023?  (and if not - what's the current canonical way to handle this?  The link from 2019 is broken and it's not obvious to me how that maps to current features.)

At this point, every supported (and recently supported) Rails version handles unscope this way.  Rails seems unlikely to change the behavior on their end - was discussed on https://github.com/rails/rails/issues/33719  I kinda see where they are coming from at that lower level context, but I think it'd be a reasonable assumption to make here to make pagy more user friendly.

Seems like it'd be straightforward enough to just not apply the .unscope(:order) if the object passed into pagy didn't respond_to that method so that we didn't break non-activerecord relations.

ddnexus:
> any chance you'd consider a patch like this in 2023?

I don't see any change for changing the reasonment.

> what's the current canonical way to handle this?

Sorry for the broken link. Here's the right one:
[Custom count to custom scopes](https://ddnexus.github.io/pagy/docs/how-to/#custom-count-for-custom-scopes)

> Seems like it'd be straightforward enough to just not apply the .unscope(:order) if the object passed into pagy didn't respond_to that method so that we didn't break non-activerecord relations.

I don't think pagy should fix in its own code a specific framework problem that the framework maintainers are not going to fix themselves.

PR #192, opened 2019-10-07
[WIP] Hanami extra

katafrakt:
More than a year after writing my blog post about integrating Hanami and Pagy I finally found some time to write a proper (?) Pagy extra for Hanami :wink: This obviously lacks documentation or tests. For I'm just creating a PR to discuss if my approach is ok and in line with Pagy's standards. I'll add docs and tests later.

[Here](https://github.com/hanami/bookshelf/compare/master...katafrakt:pagy-ext?expand=1)'s an example of how it looks when put into use.

### DIFF
--- a/lib/pagy/extras/hanami.rb
+++ b/lib/pagy/extras/hanami.rb
@@ -1,16 +1,4 @@
 class Pagy
-  module Hanami
-    module RepositoryIntegration
-      def page(offset, size)
-        root.offset(offset).limit(size).to_a
-      end
-    
-      def count
-        root.count
-      end
-    end
-  end
-
   module Backend ; private
     def pagy_hanami(repository, vars={})
       pagy = Pagy.new(pagy_hanami_get_vars(repository, vars))
@@ -29,11 +17,20 @@ class Pagy
     end
   end
 
+  module Helpers
+    alias_method :pagy_url_for_without_hanami, :pagy_url_for
+    def pagy_url_for_with_hanami(page, pagy)
+      _pagy_data = send(self.class.pagy_data_exposure)
+      options = { _pagy_data.vars[:page_param] => page }.merge(_pagy_data.vars[:params])
+      routes.path(self.class.pagy_routes_path, options)
+    end
+    alias_method :pagy_url_for, :pagy_url_for_with_hanami
+  end
+
   module Frontend
     module Hanami
       def self.included(base)
         base.send(:include, Pagy::Frontend)
-        base.send(:include, InstanceMethods)
         base.send(:extend, ClassMethods)
       end
 
@@ -52,22 +49,6 @@ class Pagy
           else
             @_pagy_data_exposure || :pagy_data
           end
-        end
-      end
-
-      module InstanceMethods
-        def pagy_nav(*args)
-          _raw super
-        end
-    
-        def pagy_info(*args)
-          _raw super
-        end
-
-        def pagy_url_for(page, pagy)
-          _pagy_data = send(self.class.pagy_data_exposure)
-          options = { _pagy_data.vars[:page_param] => page }.merge(_pagy_data.vars[:params])
-          routes.path(self.class.pagy_routes_path, options)
         end
       end
     end
--- a/lib/pagy/extras/hanami.rb
+++ b/lib/pagy/extras/hanami.rb
@@ -1,0 +1,75 @@
+class Pagy
+  module Hanami
+    module RepositoryIntegration
+      def page(offset, size)
+        root.offset(offset).limit(size).to_a
+      end
+    
+      def count
+        root.count
+      end
+    end
+  end
+
+  module Backend ; private
+    def pagy_hanami(repository, vars={})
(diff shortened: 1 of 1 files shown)


ddnexus:
@katafrakt thank you for this PR!

I will have to get into Hanami a bit before being able to evaluate/discuss your PR, but I will definitely do it ASAP.

ddnexus:
Hi @katafrakt 

The pagy extensions tend to be very small and don't use modules and inclusions: they just reopen the pagy classes or modules and add their own methods to them.

It looks like you wrote the alternative 3 backend methods: you may just want to put that directly into the `Pagy::Backend`. Look at other backend extensions for some example.

Then you have pagy_nav and pagy_info that don't seem to do anything, so the original pagy methods should probably work in Hanami as well. I would leave the user adding the _raw to the method right in their code.

Then I see that you have a lot of code around the `pagy_get_url` method. Does the original method work as is with Hanami or not? And if not... why not?

Thanks

katafrakt:
Thanks for review, @ddnexus!

> It looks like you wrote the alternative 3 backend methods: you may just want to put that directly into the Pagy::Backend

`pagy_hanami`, `pagy_hanami_get_items` and `pagy_hanami_get_items` are directly in `Pagy::Backend`. The only additional thing is a repository integration (`Pagy::Hanami::RepositoryIntegration`), which defines basic methods for repositories. Do you think that extension should leave this part totally to the programmer? I guess in many cases it would be so - there will be a need to, for example, scope only to published articles, but I wanted to add basic minimum to help people get started. If you think it's not a good idea, I will remove it.

> Then you have pagy_nav and pagy_info that don't seem to do anything, so the original pagy methods should probably work in Hanami as well. I would leave the user adding the _raw to the method right in their code.

Good idea.

> Does the original method work as is with Hanami or not? And if not... why not?

Original method relies on presence of `request` variable/method. It's not available in Hanami and it's specifically forbidden to expose it from controller to template (which is correct IMHO).

ddnexus:
Thank you for the follow up, @katafrakt.

> `Pagy::Hanami::RepositoryIntegration` [...] If you think it's not a good idea, I will remove it.

Everything that has to do with storage directly should not be in pagy. The basic principle is that your app knows how to retrieve the "thing" that has to be paginated. Pagy should just take that and paginate it with the `pagy_*get_items` method that should be the only part of the gem that deals with the actual data/scope/collection pagination.

> Original method relies on presence of request variable/method. It's not available in Hanami and it's specifically forbidden to expose it from controller to template (which is correct IMHO).

Then it makes sense to override the `pagy_url_for`. The usual way it's done in other extras is with alias chains defined directly in the re-opened module (without inclusions for simplicity, and because you need a module only if you have to include it in multiple modules). There are cleaner solutions with modern rubies, but pagy keeps compatibility with ruby 1.9, so that is the reason of alias chains. You can see a simple example in the [trim extra code](https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/trim.rb).

In your case the module to re-open would be `Pagy::Helpers` and the chain should be with/without hanami. The chaining would allow your overridden method or other extras or the developer to use the original pagy method by calling `pagy_url_for_without_hanami`. It might not be always necessary, but it is usually a good practice to let that possibility open. 

It looks like in your specific case the original method would never work under Hanami, so it might be tempting to just obliterate the original one, but you never know what some cowboy coder may come up with, so letting the original method accessible would not hurt.

katafrakt:
@ddnexus I rewrote it using `alias_method`, although original method is not called here so it kind of breaks the chain. Not sure if this is what you meant.

ddnexus:
Hey @katafrakt,
I thought I could get some time to properly learn Hanami and be useful for this extra, but waiting and waiting to find enough time is just just slowing you down. That's not good.

You could go ahead and when you say it's ready I could merge it, but I am reluctant to do that without actually knowing what I'm doing and how to properly maintain it with the rest of the project.

So I think that a lot better option would be that you move the extra into your own project that I could link in the [Related Projects](https://github.com/ddnexus/pagy#related-projects) README section (similar to what we do with [pagy-cursor](https://github.com/Uysim/pagy-cursor)).

That way the project will be decoupled from pagy and will give you full control, merit and responsibilities that I am currently not able to take. Off course I could help if you need anything about pagy and its integration with your extra, but you should be in charge with the Hanami extra.

Hoping to find you in agreement on that.

On the technical side:

> I rewrote it using `alias_method`, although original method is not called here so it kind of breaks the chain.

Even if you don't use the chaining in the extra itself it allows the user to call the original method by calling :pagy_url_for_without_hanami.

From a formal/integration point of view the code looks good now.

katafrakt:
@ddnexus That's fair. Given amount of customizations needed to get Hanami extension running, I have already thought about putting it in a separate gem. If you agree it's a good idea then we are on the same page. I will try to find some time to do that soon. I will post the link here and then we can close this PR.

ddnexus:
Let me know if you need any help on the pagy side. Please, when you will have the project to link, create a PR with the link in the README file. Thanks.

PR #196, opened 2019-10-10
Made options optional if vars param is populated accordingly

patriotphoenix:
Made options optional if vars param is populated accordingly

### DIFF
--- a/lib/pagy/extras/searchkick.rb
+++ b/lib/pagy/extras/searchkick.rb
@@ -11,9 +11,9 @@ class Pagy
 
   # create a Pagy object from a Searchkick::Results object
   def self.new_from_searchkick(results, vars={})
-    vars[:items] = results.options[:per_page]
-    vars[:page]  = results.options[:page]
-    vars[:count] = results.total_count
+    vars[:items] ||= results.options[:per_page]
+    vars[:page]  ||= results.options[:page]
+    vars[:count] ||= results.total_count
     new(vars)
   end
 
--- a/lib/pagy/extras/searchkick.rb
+++ b/lib/pagy/extras/searchkick.rb
@@ -11,9 +11,11 @@ class Pagy
 
   # create a Pagy object from a Searchkick::Results object
   def self.new_from_searchkick(results, vars={})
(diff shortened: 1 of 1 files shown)


ddnexus:
Hi @amerlescucodez 
thanks for the PR. Not sure about what would be the advantage of the changes.
The `new_from_searchkick` method is supposed to extract a pagy object out of the results: if there is no result you shouldn't even use it.
Same concept about the variable that must be in the object IMO. 
What is your use case?

amerlescucodez:
Hey ddnexus, 

I am using pagy and ran into this issue with Searchkick's multisearch. The result object when you invoke `.execute` is a Searchkick::Result, however pagy interprets that stored object as nil. Oddly enough though the `@pagy` object populates correctly with `Pagy.new_from_searchkick`, but only with this patch, otherwise it throws an exception that `result` is nil. Additionally, since the function accepts a hash of options, even when the hash of options is supplied, the existing functionality does not bypass the `result.nil?` exception.

Again, its optional and would only be used with implementing multisearch. For some reason when I wasn't using multisearch it worked fine. 

Hope that helps,

ddnexus:
Thanks for the clarification, which I am not so sure to have understood, so here are more questions:

> is a Searchkick::Result, however pagy interprets that stored object as nil

What is the exact exception? I don't think the result is nil. The result is probably a different kind of object. The options ar nil or the ones that pagy tries to fetch are nil. What is the structure of the object?

The fact that  it works with your changes is probably because it can get the count from the object and with it it creates a pagy object, which is not granted to be consistent with the other options. It probably just picks the defaults.

ddnexus:
@amerlescucodez Any update?

amerlescucodez:
Honestly I am unsure as to why the base didn't work, but the 2nd commit I made to the code base which can be seen [here](https://github.com/amerlescucodez/pagy/commit/9914afb6596b3f307f013b4c1d4b6fe3c4bd64aa) is what I have been using in my Gemfile and I haven't had an issues at all like I had prior. I think with the simple unless statement it will not error out if it can't access the record, but as to why it was happening, I am unsure. 

Hope that helps,

PR #207, opened 2019-12-05
Type cast pagy_arel_count value to integer

rainerborene:
Otherwise it will complain about "expected :count >= 0; got nil"


ddnexus:
Thank you!

Issue #219, opened 2020-02-17
Feature: Tailwind css Frontend extra


gathuku:
Requesting for a tailwind css frontend extra

ddnexus:
It looks like there is no specific pagination component for tailwind, so it's just a matter of designing your own custom helper/template.

If you want to create an extra for that, a PR would be welcome. Thanks.

gathuku:
@ddnexus  Working on it

ddnexus:
This has been replaced by #220. Closing.

PR #220, opened 2020-02-19
[WIP] Feature: Tailwindcss Frontend extra

gathuku:
This will add a tailwind css frontend extra.
https://tailwind.run/jDmZPE/3
#219

### DIFF
--- a/lib/pagy/extras/tailwindcss.rb
+++ b/lib/pagy/extras/tailwindcss.rb
@@ -4,7 +4,7 @@
 require 'pagy/extras/shared'
 
 class Pagy
-  
+
   module Frontend
     def pagy_tailwind_nav(pagy)
       link, p_prev, p_next = pagy_link_proc(pagy, 'class="block  hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2"'), pagy.prev, pagy.next
@@ -14,7 +14,7 @@ class Pagy
       pagy.series.each do |item| # series example: [1, :gap, 7, 8, "9", 10, 11, :gap, 36]
         html << if    item.is_a?(Integer); %(<li class="block hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2">#{link.call item}</li>)                                                               # page link
                 elsif item.is_a?(String) ; %(<li class="block text-white bg-blue-400 border-r border-blue px-3 py-2">#{link.call item}</li>)                                                        # active page
-                elsif item == :gap       ; %(<li class="block hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2 cursor-not-allowed"><a href="#">#{pagy_t('pagy.nav.gap'),'aria-label=". . ."'}</a></li>) # page gap
+                elsif item == :gap       ; %(<li class="block hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2 cursor-not-allowed"><a href="#">#{pagy_t('pagy.nav.gap')'}</a></li>) # page gap
                 end
       end
       html << (p_next ? %(<li class="block hover:text-blue-600 hover:bg-blue text-blue px-3 py-2">#{link.call p_next, pagy_t('pagy.nav.next'), 'aria-label="next"'}</li>)
--- a/lib/pagy/extras/tailwindcss.rb
+++ b/lib/pagy/extras/tailwindcss.rb
@@ -14,7 +14,7 @@ class Pagy
       pagy.series.each do |item| # series example: [1, :gap, 7, 8, "9", 10, 11, :gap, 36]
         html << if    item.is_a?(Integer); %(<li class="block hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2">#{link.call item}</li>)                                                               # page link
                 elsif item.is_a?(String) ; %(<li class="block text-white bg-blue-400 border-r border-blue px-3 py-2">#{link.call item}</li>)                                                        # active page
-                elsif item == :gap       ; %(<li class="block hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2 cursor-not-allowed"><a href="#">#{pagy_t('pagy.nav.gap')'}</a></li>) # page gap
+                elsif item == :gap       ; %(<li class="block hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2 cursor-not-allowed"><a href="#">#{pagy_t('pagy.nav.gap')}</a></li>) # page gap
                 end
       end
       html << (p_next ? %(<li class="block hover:text-blue-600 hover:bg-blue text-blue px-3 py-2">#{link.call p_next, pagy_t('pagy.nav.next'), 'aria-label="next"'}</li>)
--- a/lib/pagy/extras/tailwindcss.rb
+++ b/lib/pagy/extras/tailwindcss.rb
@@ -1,0 +1,26 @@
+
+# frozen_string_literal: true
+
+require 'pagy/extras/shared'
+
+class Pagy
+  
+  module Frontend
+    def pagy_tailwind_nav(pagy)
+      link, p_prev, p_next = pagy_link_proc(pagy, 'class="block  hover:text-white hover:bg-blue-600 text-blue border-r border-blue-400 px-3 py-2"'), pagy.prev, pagy.next
+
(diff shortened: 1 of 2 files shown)


ddnexus:
@gathuku How is this going?

gathuku:
@ddnexus please check this before we go to `nav-js` and `combo-nav-js`

ddnexus:
Sorry or the delay. 

Tailwingcss is a great thing and I was totally ignorant about it, so I waited to have some time to research a bit before expressing any judgment. At this point I am just a-couple-of-hours-new to tailwind so I might be wrong, but I am not sure this is the right way to use pagy with it.

I mean, the extras is the standard way pagy uses to add functionality, which is fine with any framework and it looks like you did it properly, but - if I got the basic concepts in tailwind right - shouldn't we create the tailwind styling to match with the Pagy gem's HTML and classes instead?

I mean, I think we need just a css snippet that the users can copy from this repo and paste into their `application.css`, defining the pagy classes with the `@apply`ed tailwing classes.
 
[Here is an example](https://www.youtube.com/watch?v=IKjttXqtVV0&feature=youtu.be&t=586) of what I mean. The example uses the `.btn*` classes, but imagine we define the classes that  pagy already uses in its basic helper and templates, by using `@apply` in each class.

That way instead of creating and extra with helpers and templates with hard-coded tailwind classes to the pagy items, we could apply the tailwing classes to the pagy classes, assigning there whatever style we need. The added advantage would be that the customization would be totally flexible (just edit the snipped after you paste it), without the need of any extra, nor the overriding of helpers or templates in case of customization.

What do you think?

ddnexus:
@gathuku did you have a chance to read my comment?

gathuku:
@ddnexus  this is a great idea. With use of tailwind  `@apply` directives it's possible to extend the existing classes to apply tailwind classes.
How will it go? Does each extra have different classes?

ddnexus:
These are the files and helpers that use CSS:

https://github.com/ddnexus/pagy/blob/master/lib/pagy/frontend.rb
pagy_nav
pagy_info

https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/navs.rb
pagy_nav_js
pagy_combo_nav_js


https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/items.rb
pagy_items_selector_js

gathuku:
@ddnexus  Just gave this a try and is working fine. I come up with below extension, it is not perfect but can give someone an idea to get started. Please recommend
```
.pagy-nav {
  @apply .inline-flex .shadow-md
}
.pagination{
  @apply .border .border-gray-600 .rounded-sm
}
.page {
  @apply  .text-gray-500 .border-r .border-gray-600 .px-3 .py-2 .text-sm .leading-tight .font-medium;
}
.page:hover {
  @apply .text-gray-900
}
.disabled {
  @apply .cursor-not-allowed
}
.active {
  @apply .text-blue-500 .leading-none
}
.prev {
  @apply .text-gray-900
}
.next {
  @apply .text-gray-900 .border-r .border-transparent
}
```

ddnexus:
That looks like a good start! 
I don't have any app with tailwind to try it. Please, could you post a screenshot?

As I said, I am quite new to tailwind, so I might ask stupid questions... is there a way to assign a color based on some conventional variable, instead of hard-coding blue and gray?
I mean some color-variable (e.g. something like primary-color, secondary-color, background-color) that has already defaults and the users are supposed to customize in their app, so that will automatically change also the pagy item colors?

ddnexus:
And another thing: are we sure that the selectors would not pick some other class in the app?

I mean: applying directly to the .disabled, .active, .prev, .next, etc. without restricting that to the item inside the pagy nav makes me think that it could screw other app classes.

gathuku:
> That looks like a good start!
> I don't have any app with tailwind to try it. Please, could you post a screenshot?
> 
> As I said, I am quite new to tailwind, so I might ask stupid questions... is there a way to assign a color based on some conventional variable, instead of hard-coding blue and gray?
> I mean some theme or global color-variable (e.g. something like primary-color, secondary-color, background-color) that has already defaults and the users are supposed to customize in their app, so that will automatically change also the pagy item colors?

 
Screenshot from 2020-04-19 21 38 01
Screenshot from 2020-04-19 21 39 09

@ddnexus  Yes , you can set up some global themes with css variables in `tailwindcss.config.js` , Please see more https://www.samdawson.dev/article/tailwind-theming

gathuku:
> And another thing: are we sure that the selectors would not pick some other class in the app?
> 
> I mean: applying directly to the .disabled, .active, .prev, .next, etc. without restricting that to the item inside the pagy nav makes me think that it could screw other app classes.

This true it can mess with other classes. Need to be restricted

ddnexus:
> Yes , you can set up some global themes with css variables in tailwindcss.config.js

I was referring to possible default themes or variables or anything conventionally **already** defined in tailwind defaults. If there is any of that, we should use it, otherwise we should not bother, because it will be up to the user.

> This true it can mess with other classes. Need to be restricted

So let's create pagy specific selectors

JeremiahChurch:
CSS descendants should allow properly containing all of the pagination options within the existing pagination nav classes. At least as far as the `pagy_nav` goes. Working from @gathuku 's examples - just adding .pagy-nav as the descendant css class should cleanly restrict everything so that no @applies are overly aggressive

```scss
.pagy-nav {
  @apply .inline-flex .shadow-md;
}
.pagy-nav.pagination {
  @apply .border .border-gray-600 .rounded-sm;
}
.pagy-nav .page {
  @apply .text-gray-500 .border-r .border-gray-600 .px-3 .py-2 .text-sm .leading-tight .font-medium;
}
.pagy-nav .page:hover {
  @apply .text-gray-900;
}
.pagy-nav .disabled {
  @apply .cursor-not-allowed;
}
.pagy-nav .active {
  @apply .text-blue-500 .leading-none;
}
.pagy-nav .prev {
  @apply .text-gray-900;
}
.pagy-nav .next {
  @apply .text-gray-900 .border-r .border-transparent;
}
```

ddnexus:
@JeremiahChurch @gathuku I guess Tailwind does not support nested classes SASS style, right?

JeremiahChurch:
@ddnexus tailwind comes at the problem entirely different than most of the other popular style frameworks by going with utility classes for everything instead of component classes. So you end up needing to create your own component classes or partials. Which is where in my experience you see nested classes get used.

It didn't really click for me until I saw their button example [here](https://tailwindcss.com/docs/extracting-components#extracting-css-components-with-apply). Nested sass classes wouldn't cause any problems - but they'd end up outside of the default tailwind utility classes and would need to be associated with them via @apply or similar.

ddnexus:
@JeremiahChurch Thank for the explanation: ! have to trust you on this one because I guess I will have to find some time to play with tailwind before I could really say that anything really "clicks" :)

I think he `pagy_combo_nav_js` has some specific class missing from the list before considering it complete though.

gathuku:
> CSS descendants should allow properly containing all of the pagination options within the existing pagination nav classes. At least as far as the `pagy_nav` goes. Working from @gathuku 's examples - just adding .pagy-nav as the descendant css class should cleanly restrict everything so that no @applies are overly aggressive
> 
> ```css-scss
> .pagy-nav {
>   @apply .inline-flex .shadow-md;
> }
> .pagy-nav.pagination {
>   @apply .border .border-gray-600 .rounded-sm;
> }
> .pagy-nav .page {
>   @apply .text-gray-500 .border-r .border-gray-600 .px-3 .py-2 .text-sm .leading-tight .font-medium;
> }
> .pagy-nav .page:hover {
>   @apply .text-gray-900;
> }
> .pagy-nav .disabled {
>   @apply .cursor-not-allowed;
> }
> .pagy-nav .active {
>   @apply .text-blue-500 .leading-none;
> }
> .pagy-nav .prev {
>   @apply .text-gray-900;
> }
> .pagy-nav .next {
>   @apply .text-gray-900 .border-r .border-transparent;
> }
> ```
@JeremiahChurch  this is a great idea.

gathuku:
@ddnexus  does `pagy_combo_nav_js` applies to extras only? is it available in the default frontend?

ddnexus:
All the frontend extras style pagy_nav, pagy_nav_js and pagy_combo_nav_js, so adding the styles for the missing items should be standard.

ddnexus:
Added extra styles in doc 62bd4acb

minimul:
I'm running 3.8.3 and `extras/tailwind` is not in the release. I checked the local gem source code too.

ddnexus:
@minimul it's actually not a require-able extra, since Tailwind doesn't need anything from ruby. It's just a complete set of rules that you can copy and customize in your app. 
Since it was not obvious where to actually put the rules in order to be useful to the users, I just added it as a [Tailwind page in the doc](https://ddnexus.github.io/pagy/extras/tailwind).
HTH

PR #226, opened 2020-03-17
Rename Swedish locale to sv.yml

davidwessman:
- The Swedish locale is referred to as sv in the I18n-gem:
  https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/sv.yml
- It can also be referred to as sv_SE.

### DIFF
--- a/lib/locales/sv-SE.yml
+++ b/lib/locales/sv-SE.yml
@@ -1,0 +1,23 @@
+# :one_two_other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+
+sv-SE:
+  pagy:
+
+    item_name:
+      one: "resultat"
+      two: "resultat"
+      other: "resultat"
+
+    nav:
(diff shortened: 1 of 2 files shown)


davidwessman:
I decided to rename it completely, but feel free to keep the old file for backwards compatibility if that is needed.

ddnexus:
I think we should have a `sv.yml` containing the `sv` key and a `sv-SE.yml` containing `sv-SE`,  since they are also in the official i18n gem. We used the same file/key naming for the `zn-CH`, `zn-HK` and the `zn-TW`.

Please, could you modify the PR accordingly?

davidwessman:
> I think we should have a `sv.yml` containing the `sv` key and a `sv-SE.yml` containing `sv-SE`, since they are also in the official i18n gem. We used the same file/key naming for the `zn-CH`, `zn-HK` and the `zn-TW`.
> 
> Please, could you modify the PR accordingly?

Done!

ddnexus:
Thank you. I will add the needed changes for pluralization and deprecation.

PR #228, opened 2020-03-20
add items_per_page set to original value of items

jellymann:
I ran into an astonishing issue. I was passing the Pagy instance to my custom pagination component and using its `items` attribute to set the selected option of a dropdown to select items per page. I happened to filter my data so that there were only 5 results, and the select changed to 5. I then discovered that `items` gets adjusted to the number of items when on the last page (when there were only 5 results, there was only one page, so it was on the last page).

As a workaround, I've had to explicitly pass the `items` value separately to my pagination component. However it would be nice if Pagy kept the original value of `items` somewhere so that it can be used for purposes similar to my own.

This PR is my suggestion as to how this could be achieved without any breaking changes.

### DIFF
--- a/lib/pagy.rb
+++ b/lib/pagy.rb
@@ -12,7 +12,7 @@ class Pagy ; VERSION = '3.7.3'
   # default vars
   VARS = { page:1, items:20, outset:0, size:[1,4,4,1], page_param: :page, params:{}, anchor:'', link_extra:'', i18n_key:'pagy.item_name', cycle:false }
 
-  attr_reader :count, :page, :items, :vars, :pages, :last, :offset, :from, :to, :prev, :next
+  attr_reader :count, :page, :items, :vars, :pages, :last, :offset, :from, :to, :prev, :next, :items_per_page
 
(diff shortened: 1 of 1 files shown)


ddnexus:
@jellymann thanks for the PR, but I don't see the need for an extra variable to save the original items number. Pagy already stores it in the `vars`:

```irb
>> pagy = Pagy.new(count: 5)
#<Pagy:0x000055823a7fae10 @vars={:page=>1, :items=>20, :outset=>0, :size=>[1, 4, 4, 1], :page_param=>:page, :params=>{}, :anchor=>"", :link_extra=>"", :i18n_key=>"pagy.item_name", :cycle=>false, :count=>5}, @count=5, @items=5, @outset=0, @page=1, @last=1, @pages=1, @offset=0, @from=1, @to=5, @prev=nil, @next=nil>
>> pagy.items
5
>> pagy.vars[:items]
20
```

jellymann:
@ddnexus Oh my gosh I didn't even realise that 🤦‍♂ 

Thanks for pointing it out. 👍

PR #230, opened 2020-03-24
Reduce DB queries by replacing .count(:all) with .size

duartemvix:
Made a few changes to provide Pagy::Backend module with AR Relations. These relations should already be loaded in Rails controllers before calling methods from the module.

This will remove an extra COUNT query made to databases and sort the relation of records as an Array inside the module.

Produce an AR Relation inside controllers and apply any required scopes by

```ruby
class RelationsController < ApplicationController
  def index
    @relation = Relation.includes(:model).scope.desired.by_time
    @relation.load
    @pagy, @relation = pagy @relation
  end
end
```

### DIFF
--- a/lib/pagy/backend.rb
+++ b/lib/pagy/backend.rb
@@ -18,7 +18,7 @@ class Pagy
 
     # Sub-method called only by #pagy: here for easy customization of variables by overriding
     def pagy_get_vars(collection, vars)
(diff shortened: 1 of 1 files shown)


duartemvix:
Looking closely, this code would load too much data into memory if you're running large databases. I decided to skip this and go back but I'll leave it here in case somebody wants to test it.

ddnexus:
Yes, but good to see that you tried! Thank you anyway.

PR #241, opened 2020-05-08
fix for deprecation warnings in ruby 2.7.x

majksner:
This pull request fixes the warnings for Ruby 2.7.x.

### DIFF
--- a/lib/pagy/extras/i18n.rb
+++ b/lib/pagy/extras/i18n.rb
@@ -11,7 +11,7 @@ class Pagy
     Pagy::I18n.clear.instance_eval { undef :load; undef :t } # unload the pagy default constant for efficiency
 
     alias_method :pagy_without_i18n, :pagy_t
-    def pagy_t_with_i18n(*args) ::I18n.t(*args) end
+    def pagy_t_with_i18n(**args) ::I18n.t(*args) end
     alias_method :pagy_t, :pagy_t_with_i18n
 
   end
--- a/lib/pagy/extras/shared.rb
+++ b/lib/pagy/extras/shared.rb
@@ -27,13 +27,13 @@ class Pagy
 
     if defined?(Oj)
       # it returns a script tag with the JSON-serialized args generated with the faster oj gem
-      def pagy_json_tag(*args)
+      def pagy_json_tag(**args)
         %(<script type="application/json" class="pagy-json">#{Oj.dump(args, mode: :strict)}</script>)
       end
     else
       require 'json'
       # it returns a script tag with the JSON-serialized args generated with the slower to_json
-      def pagy_json_tag(*args)
+      def pagy_json_tag(**args)
         %(<script type="application/json" class="pagy-json">#{args.to_json}</script>)
       end
     end


ddnexus:
Doesn't that fail with 1.9 syntax?

majksner:
Isn't the ruby 1.9 end of life?

ddnexus:
Regardless the end of life of 1.9, I wouldn't like to break systems that still use it, especially because pagy is backward compatible till 1.9 and I personally use it in legacy systems. 

Besides, I just noticed that your change does actually change the semantic of the method, and does not work as it is supposed to work. `def meth(*args) ...` catches all the passed arguments in array, while `def meth(**args)` defines something completely different.

Indeed i18n tests fail with your change.

ddnexus:
And... I don't see any deprecation warning running the tests in 2.7.1 (https://travis-ci.org/github/ddnexus/pagy/jobs/678434972).

Where is your warning coming from?

majksner:
```
/Users/nikola/.rvm/gems/ruby-2.7.1/gems/pagy-3.8.0/lib/pagy/extras/i18n.rb:14: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/nikola/.rvm/gems/ruby-2.7.1/gems/i18n-1.8.2/lib/i18n.rb:195: warning: The called method `t' is defined here
```

majksner:
Seems it comes from https://github.com/ruby-i18n/i18n/

I'm sorry.

ddnexus:
It actually looks like it may come from how pagy calls the `t` method. The tests don't raise any warning, but their call might be different than in the actual call made in a page. I will check that. Thanks.

ddnexus:
It could be due to the i18n 1.8.2 version. Travis runs with 1.6.0. Please could you temporarily downgrade it to 1.6.0 and check if the warning is still there?

majksner:
The warning is not displayed with i18n 1.6.0.

Check this out: https://github.com/ruby-i18n/i18n/blob/master/lib/i18n.rb#L188

ddnexus:
Your link points to the 1.8.2 code. 

The [1.6.0 code](https://github.com/ruby-i18n/i18n/blob/529b83a32036c210e4e35d49a2ac23e956cca67a/lib/i18n.rb#L179) defines the same arguments for the method I really don't get how it could not raise the same warning if you still run it on 2.7+. 

The only difference about the versions is a comment added to the method (see https://github.com/ruby-i18n/i18n/pull/501), so I really don't get how it could change anything with the warning.

ddnexus:
it breaks old rubies though :/

majksner:
Maybe something like this can be used?

```ruby
if RUBY_VERSION >= "2.7"
  def pagy_t_with_i18n(key, **opts) ::I18n.t(key, **opts) end
else
  def pagy_t_with_i18n(*args) ::I18n.t(*args) end
end
```

ddnexus:
Thank you for the effort, but that would not avoid the syntax error for ruby <2.0. 
The above commit does it.

ddnexus:
BTW, `RUBY_VERSION >= "2.7"` does not do what you think. :)

ddnexus:
[Indeed!](https://travis-ci.org/github/ddnexus/pagy/builds/684646275)

PR #244, opened 2020-05-30
Use proper attributes for "previous" links

marckohlbrugge:
Changes made:

- the `rel` attribute has been changed from `next` to `prev`
- the `aria-label` attribute has been changed from `next` to `previous`
- the associated tests have been updated

I chose not to abbreviate the aria label as it's intended for screen readers, etc. In fact, `aria-labelledby` might be a better alternative as we already provide a description within the anchor text itself. But I decided not to address that issue in this PR.

### DIFF
--- a/lib/pagy/extras/support.rb
+++ b/lib/pagy/extras/support.rb
@@ -15,7 +15,7 @@ class Pagy
     end
 
     def pagy_prev_link(pagy, text = pagy_t('pagy.nav.prev'), link_extra = '')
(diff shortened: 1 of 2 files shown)


ddnexus:
Thank you!

Feel free to create a very welcomed PR also for the fix and improvement of the aria attributes.

marckohlbrugge:
Should we merge this PR first, and take care of the aria attributes in a second PR? I'd like to read up on aria first just to make sure I got it right.

ddnexus:
> I'd like to read up on aria first just to make sure I got it right.

Sure. If you decide to improve it, please, link also some reference.

PR #247, opened 2020-06-27
Add class option for bulma extra

ippachi:
I use this gem with bulma.
And I want to `is-rounded` option but It doesn't seem to have nav class customizations.

Thank you for the great gem!


ddnexus:
Thanks for the PR. It looks good but it adds an argument to the helper just to pass a CSS class, and that has been avoided till now.

Adding a CSS class to an element is quite easy. You can use: 

- the `@extend` sass/scss directive when available
- the jQuery `addClass` method
- a couple of lines of plain javascript

If none of the above is possible, pagy makes the overriding of helpers or templates in your own code very easy as well.

So while your PR makes perfect sense from the point of view of your own specific need, it complicates things from the general point of view of the whole pagy gem policies and maintenance. That's the reason I would prefer not merging it.

ippachi:
Thank you for the explanation and for your help.
I understood.
I’ll use javascript 😄

PR #250, opened 2020-07-08
Add Portuguese translation

tiagotex:
The translation for Portuguese from Brazil was already present but we were missing the translation for Portuguese from Portugal.

Let me know if any other changes are required.


ddnexus:
Thank you!

PR #258, opened 2020-08-31
Remove ordering for collection before getting count

natesalisbury:
When a order is present it can break count calls. See: https://github.com/rails/rails/issues/33719.

I'm not sure if this is too specific to ActiveRecord for this class. But I have a handful of places I need to manually remove the order and pass in the count to avoid an error.

### DIFF
--- a/lib/pagy/backend.rb
+++ b/lib/pagy/backend.rb
@@ -18,7 +18,10 @@ class Pagy
 
     # Sub-method called only by #pagy: here for easy customization of variables by overriding
     def pagy_get_vars(collection, vars)
-      vars[:count] ||= (c = collection.count(:all)).is_a?(Hash) ? c.size : c
+      unless vars[:count]
+        collection = collection.reorder(nil) if collection.respond_to?(:reorder)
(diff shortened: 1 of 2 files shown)


ddnexus:
Thank you for your PR and sorry for being late to comment. 
There are pros and cons to what you are proposing. I will post as soon as I will have a bit of time. Thank you.

deepj:
I guess, the better option would be some configurable settings what would call `except` to remove some conditions by default. In my case, it is typical `eager_load` what generates expensive queries with `DISTINCT` when some of eager loaded associations is `has_many`. In your example it would be something like `query.except(:order)`

ddnexus:
That is very specific code for a general purpose pagination gem. 
We had the same kind of dilemma for the [arel extra](https://ddnexus.github.io/pagy/extras/arel), which decided to put in its own extra because it is dealing with a specific situation/enhancement.

Maybe we should extend that extra or merge it into an activerecord extra that will be specifically dealing with the AR quirks.
What do you think?

ddnexus:
Closing for inactivity

PR #259, opened 2020-09-18
Add: tutorial to readme

BKSpurgeon:
A simple tutorial on how to make those pagination links sticky + making them "bounce" up when you reach the bottom. It may benefit other pagy users.

Feel free to discard the PR if it doesn't add value.

rgds

Ben


ddnexus:
It's not pagy specific, but it's a nice short tutorial that could be useful to someone.

PR #264, opened 2020-10-15
Update pagy.rb to handle asking for a page > the last page

weiserma:
raising an error when what was a good page number is no longer good if a waste of code! Exceptions should only be raised for exceptional conditions no standard processing.

when i first view the page there are 20 tasks... someone else bulk closes 15... now there is only 1 page and a refresh needs to by exception handed.... NO WAY!


ddnexus:
With your change we would impose a behavior that is not compatible with certain API clients (that stop fetching pages only when they get an empty page). And what if a user would actually want to return page 1 or do anything else instead of getting the last page? 

The [Overflow extra](https://ddnexus.github.io/pagy/extras/overflow) handles a few common cases (including `:last_page`), it is compatible also with the countless extra, and does not impose any restriction on the user.

so... "NO WAY!" :smile: 
Thank you

PR #279, opened 2021-03-03
Ruby 3 compatibility for Pagy::Search and Searchkick extra

tosch:
When using Pagy together with Searchkick in Ruby 3, there is an ArgumentError raised. This is because Pagy::Search has been using an options hash, while Searchkick expects keyword arguments to be used on the method call. In previous Ruby versions, Ruby did some magic to auto-convert the last argument to keyword args if it is an Hash. This is no longer the case in Ruby 3.

The failure can be seen in this minimal Rails application: https://github.com/tosch/pagy-searchkick-minimal-rails

This fix captures any keyword arguments given to Pagy::Search#pagy_search and ensures they are being used correctly when either using Elasticsearch (which expects an options hash as argument) or using Searchkick (which accepts an optional search term and keyword arguments).

With those changes applied, the above minimal Rails application works as expected.

### DIFF
--- a/lib/pagy/extras/searchkick.rb
+++ b/lib/pagy/extras/searchkick.rb
@@ -26,7 +26,7 @@ class Pagy
       vars                       = pagy_searchkick_get_vars(nil, vars)
       search_args[-1][:per_page] = vars[:items]
       search_args[-1][:page]     = vars[:page]
-      results                    = model.search(*search_args, &block)
+      results                    = pagy_searchkick_get_results(model, search_args, block)
       vars[:count]               = results.total_count
       pagy = Pagy.new(vars)
       # with :last_page overflow we need to re-run the method in order to get the hits
@@ -42,6 +42,19 @@ class Pagy
       vars[:items] ||= VARS[:items]
       vars[:page]  ||= (params[ vars[:page_param] || VARS[:page_param] ] || 1).to_i
       vars
+    end
+
+    if RUBY_VERSION.start_with?('3.')
+      eval <<-RUBY
+        def pagy_searchkick_get_results(model, search_args, block)
+          options = search_args.pop
+          model.search(*search_args, **options, &block)
+        end
+      RUBY
+    else
+      def pagy_searchkick_get_results(model, search_args, block)
+        model.search(*search_args, &block)
+      end
     end
 
   end


ddnexus:
Thanks for the contribution!

However there is a problem with the double splat syntax. Pagy has been compatible with ruby 1.9+ since the beginning, and the double splat has been introduced with 2.0. Accepting the contribution as-is could break systems that runs on 1.9 or jruby 1.7+.

Could you find a way to avoid the double splat?

If not the only alternative seems to be dropping the support for ruby <2 and create a pagy v4, which I would like to avoid for now.

I saw that you changed the expected result in the test. I am not so familiar with the result object of both elasticsearch gems, so if you could give me some clue about that, I would save lot of time.

Thanks

tosch:
Yeah, I came to the same conclusion about the missing Ruby <2 support yesterday evening. I'm not sure if it will be possible to change that, but maybe I can try something Ruby version dependent? I'll see if I can come up with something.

The changes of the result objects in the tests are not the actual responses from Elasticsearch, but those from Pagy::Search#pagy_search. The array that is returned from that method now has four elements instead of the previous three so that the options can be passed on separately from the arguments.

ddnexus:
I had to avoid the double splat in  couple of occasion with pagy, and it has not been difficult, just more verbose. Hoping it will be the same for this PR.
Thanks for the clue about the test change: I didn't have much time to check it well and it makes sense now.

tosch:
I've tried a few things and I'm not too enthusiastic about the results. The main issue is [this line](https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/searchkick.rb#L29) in the searchkick extra. For Ruby 2, the options hash is converted into keyword arguments without issues. In Ruby 3, this is no longer the case and an argument error is raised, because Searchkick::Model#search only expects one argument, but Pagy::Backend#pagy_searchkick is passing two arguments, with the last one being the options hash.

Now, I thought I could move the call to the search method into another private method and define that one differently based on the Ruby version that is run:

```ruby
    if RUBY_VERSION.start_with?('3.')
      def pagy_searchkick_get_results(model, search_args, block)
        options = search_args.pop
        model.search(*search_args, **options, &block)
      end
    else
      def pagy_searchkick_get_results(model, search_args, block)
        model.search(*search_args, &block)
      end
    end
```

That doesn't work, though, as the syntax is checked on the whole file and it's no correct syntax in Ruby < 2 because of the double splat operator. I don't think there is another way to pass keyword arguments to methods, at least I didn't find one. So those are the options I see, what do you think about them?

1. Use the Ruby version specific code, but put the Ruby >= 3 code into an eval'ed string:

   ```ruby
    if RUBY_VERSION.start_with?('3.')
      eval <<-RUBY
        def pagy_searchkick_get_results(model, search_args, block)
          options = search_args.pop
          model.search(*search_args, **options, &block)
        end
      RUBY
    else
      def pagy_searchkick_get_results(model, search_args, block)
        model.search(*search_args, &block)
      end
    end
    ```
 
 2. Put the offending code into Ruby version specific separate files and only require the file based on the major Ruby version. This would avoid the `eval`.
 
 3. Drop support for Ruby 1.x and release v4 of Pagy, so that keyword arguments can be used without issues.
 
What's your preference? Option 1 fixes the issue in the minimal Rails app I've mentioned, so besides of using evil eval, that might be a fix.

tosch:
I'd really like to increase the coverage on this PR, but that would require running the tests in Ruby 3 as well — I didn't have time to look into how to set this up.

tosch:
Maybe worth referencing the article that describes the changes in Ruby that cause this issue: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

ddnexus:
Thank you for the time you are dedicating to this.

> Searchkick::Model#search only expects one argument

I am confused by that statement. In the Searchkick page I read also `Product.search("apples", load: false)` 

However, your options makes sense to me. The eval is not that bad, since it is done only once at require time. There are quite a few other evals at require time... maybe not pretty, but a lot of pagy code is not pretty but efficient and practical, so it would be OK using it.

Now, despite the fact that I would like to avoid v4 at this time, that does not necessarily mean that keeping the compatibility to 1.9+ at all costs is the way to go. Indeed it has been a restriction in the past and that will only worsen with time, so it's probably time to do drop it, also because the v3 will still work for legacy systems.

Please, let me see how I can allocate and the required time and when. After that your PR could be applied to v4 without any problems.

Please, could you stream down a PR that would work on ruby >2? I will create a v4 branch and start to merge changes and PR there.

Thanks
Thank you.

ddnexus:
@tosch I prepared the branch [ 4.0](https://github.com/ddnexus/pagy/tree/4.0). You can use `ruby 3.0+` syntax in it, since it has a required_ruby_version `>= 3.0`.

I added also a [docker development environment](https://github.com/ddnexus/pagy/tree/4.0/docker) that should ease development.

ddnexus:
Ok, please, could you try the branch '4.0' with your ruby 3.0 app? The only thing you should change is `Pagy::Search` to `Pagy::Searchkick`. Please, let me know.

tosch:
@ddnexus Wow, that's awesome, thanks for that work. Sorry for not replying during the last couple of days, I was a bit sick and unable to sit at my desk.

The test on my Ruby 3 app ran successfully:

```
$ bin/rails test
Running via Spring preloader in process 36520
Run options: --seed 23910

# Running:

.

Finished in 0.902248s, 1.1083 runs/s, 1.1083 assertions/s.
1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

So, your 4.0 branch is fixing the issue. Thanks again, your work on Pagy is much appreciated!

ddnexus:
Thank you for your pull request, raising the issue and discussing the solutions. Your contribution is very appreciated!

PR #281, opened 2021-03-11
Add translate for Bosnian and Croatian language

nedimdz:
Hi there!

Here is the Bosnian and Croatian translation. It's two different languages with small differences, but in this specific case there is the same translation for items.

I hope you can use it 👍

### DIFF
--- a/lib/locales/bs.yml
+++ b/lib/locales/bs.yml
@@ -1,0 +1,24 @@
+# :east_slavic pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+
+bs:
+  pagy:
+
+    item_name:
+      one: "stavka"
+      few: "stavke"
+      many: "stavki"
+      other: "stavki"
+
+    nav:
+      prev: "&lsaquo;&nbsp;Prethodna"
+      next: "Sljedeća &nbsp;&rsaquo;"
+      gap: "&hellip;"
+
+    info:
+      no_items: "Nije pronađena %{item_name}"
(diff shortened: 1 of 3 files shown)


ddnexus:
Thank you for the PR!

I have a couple of questions to ask you:

1. According to rails-i18n repo both languages are listed as east-slavic ([bs](https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/bs.rb) and [hr](https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/hr.rb)), while you indicate `:west-slavic` in the files. There is quite a difference in how the pluralization works for east and west, as you can see here: https://github.com/ddnexus/pagy/blob/3a0aafb3acea443abfb998537c064280c51f3943/lib/locales/utils/p11n.rb#L23-L40 Which one would be the correct one? 
2. That should be added in the list here: https://github.com/ddnexus/pagy/blob/3a0aafb3acea443abfb998537c064280c51f3943/lib/locales/utils/p11n.rb#L69-L84
Could you add both language in the pluralization list?

Thanks

nedimdz:
Thanks for the reply. To be honest, I didn’t check the rails - i18n repository, but looking at the logic of 
```
west_slavic: lambda do |n| 
    if n == 1                    ; 'one' 
    elsif [2, 3, 4].include?(n)  ; 'few' 
    else                         ; 'other' 
    end 
 end
```

It's seem ok in this case, but I will put east_slavic it is certainly more correct.

ddnexus:
Thank you!

PR #283, opened 2021-03-11
Add Serbian language translation

sabljak:
Hello.

Here's the Serbian locale.  For `no_items` the sentence will be written in a different way depending on the count of items. 
For example:

- one item:  "Није пронађенa %{item_name}"
- multiple items: "Нema пронађених %{item_name}"

Is there a way to use one or the other depending on the count?

😃

### DIFF
--- a/lib/locales/sr.yml
+++ b/lib/locales/sr.yml
@@ -1,0 +1,23 @@
+# :east_slavic pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+
+sr:
+  pagy:
+    item_name:
+      one: "ставка"
+      few: "ставкe"
+      many: "ставки"
+      other: "ставки"
+
+    nav:
+      prev: "&lsaquo;&nbsp;Претходна"
+      next: "Следећa &nbsp;&rsaquo;"
+      gap: "&hellip;"
+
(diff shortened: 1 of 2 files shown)


ddnexus:
Thank you for your PR!

> Is there a way to use one or the other depending on the count?

Yes, for the `no_items` you basically do the same as for the `item_name`, i.e. adding all the plurals for the language (even if there are duplicates). Full info [here](https://github.com/ddnexus/pagy/tree/master/lib/locales).

@nedimdz do `bs` and `hr` follow the same multiple translations for `no_items` ?

sabljak:
I hope this resolves the issue.

ddnexus:
@sabljak it does not solve the issue at all, since the `no_item` key that you added will never get used by the code. 

Actually looking at the "problem" again, I missed the fact that `no_items` get used only when the count is `0`, so I don't get how you may need different pluralizations when the count passed for `no_items` will be always `0`.

Could you explain me what you mean?

sabljak:
My bad, I didn't think it all the way through. The initial `no_items` translation will do just fine.

sabljak:
@ddnexus All done. Hope everything is OK.

ddnexus:
Thank you!

PR #285, opened 2021-03-24
Fix Tailwind docs

thomasklemm:
Tailwind's `@apply` directive works by taking the utility class name without a `.` in front. For reference, here's the [Tailwind v2 `@apply` examples](https://tailwindcss.com/docs/extracting-components#extracting-component-classes-with-apply). Same for Tailwind v1.

This PR makes the base styles proposed in the docs for Tailwind CSS users work directly on copy-pasting, doesn't change any of the styles.

Prevents the following error:
```
15:22:48 webpack.1 | ModuleBuildError: Module build failed (from ./node_modules/postcss-loader/dist/cjs.js):
15:22:48 webpack.1 | SyntaxError
15:22:48 webpack.1 |
15:22:48 webpack.1 | (4:2) /Users/thomasklemm/Code/Ashla/sunrise/app/javascript/stylesheets/components/_pagination.scss The `.inline-flex` class does not exist, but `inline-flex` does. If you're sure that `.inline-flex` exists, make sure that any `@import` statements are being properly processed before Tailwind CSS sees your CSS, as `@apply` can only be used for classes in the same CSS tree.
15:22:48 webpack.1 |
15:22:48 webpack.1 |   2 | .pagy-nav-js,
15:22:48 webpack.1 |   3 | .pagy-combo-nav-js {
15:22:48 webpack.1 | > 4 |   @apply .inline-flex .shadow-md;
15:22:48 webpack.1 |     |  ^
15:22:48 webpack.1 |   5 | }
```

### DIFF
--- a/docs/extras/tailwind.md
+++ b/docs/extras/tailwind.md
@@ -15,40 +15,40 @@ Copy and customize the following basic rules to apply the styles to the pagy CSS
 .pagy-nav, 
 .pagy-nav-js,
 .pagy-combo-nav-js {
-  @apply .inline-flex .shadow-md;
+  @apply inline-flex shadow-md;
 }
 .pagy-nav.pagination, 
 .pagy-nav-js.pagination,
 .pagy-combo-nav-js.pagination {
-  @apply .border .border-gray-600 .rounded-sm;
+  @apply border border-gray-600 rounded-sm;
 }
 .pagy-nav .page,
 .pagy-nav-js .page,
 .pagy-combo-nav-js .page,
 .pagy-combo-nav-js .pagy-combo-input {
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you for the PR!

> Same for Tailwind v1
Not really. The leading dot was supported in version `<2`, but removed in version `2` (see [here](https://tailwindcss.com/docs/upgrading-to-v2#remove-leading-dot-from-apply-statements)).

However, since it was optional, we can merge your PR because old version would work as well without the leading dot, so it will work for old and new.

Thanks

thomasklemm:
> The leading dot was supported in version <2, but removed in version 2 (see here).

@ddnexus Oh nice, I see, thanks for setting this straight.

Issue #290, opened 2021-04-07
I18n Extra no longer performs as expected in pagy >= 4.0.0


whithajess:
We have the following in our `config/initializers/pagy.rb` file:
```
# Use the standard I18n gem - avoids special cases for new translations.
#
# https://ddnexus.github.io/pagy/api/frontend#using-the-standard-i18n-gem
#
require 'pagy/extras/i18n'
```
And have had it in for all previous versions of pagy working correctly up until upgrading past `3.12.0` where we saw errors like so in tests:
```
ActionView::Template::Error:
  undefined method `t' for Pagy::Frontend::I18n:Module
```
commenting out the `require 'pagy/extras/i18n'` line fixes this.

Im assuming its related to this change: https://github.com/ddnexus/pagy/commit/f1d89a5

whithajess:
Can confirm reverting the refactoring fixes the bug.

ddnexus:
Thank you for the report. Can you take a look at why the tests don't fail?

whithajess:
^ @ddnexus its basically a namespace collision issue

ddnexus:
Thanks for the test that exposes the problem.

As you pointed out it's a namespace problem. The `I18n` module name to prepend is an homonym of the `I18n` gem namespace, and either we impose to all the user to write `::I18n` when they mean the gem namespace - which doesn't seem a good idea :) - or we can just rename the module to prepend with a sane `UseI18nGem` name, and we solve the conflict and pass the test.

If you have time, please could you fix it by changing the name, instead of reverting the commit to the ugliness of alias chaining (that we were using only to please old rubies)? If you create a new PR, please, add the new test, which is important.

Thank you

whithajess:
Fix including test pushed ^

whithajess:
@ddnexus would be great to get this merged and released.

PR #292, opened 2021-04-07
Add a test that looks at i18n working in Modules included

whithajess:
* Demonstrates https://github.com/ddnexus/pagy/issues/290


whithajess:
Failure case added to PR: https://github.com/ddnexus/pagy/pull/293

PR #296, opened 2021-04-13
CI: Remove Bundler 1.x workaround

olleolleolle:
The "github" source now has that form, already.

This PR only serves to simplify that Gemfile.


ddnexus:
Thank you!

This PR was failing in travis.ci. Then, since it's an intermittent problem happening since a few days, I just re-built it and it worked this time. 

I think it's related to the warning of codecov circular require, which is happening because simplecov requires it on its own with docile when on travis and on ruby 3.0, and there is a require in the `.simplecov` file too. I will fix that next ;).

Thanks

PR #297, opened 2021-04-13
gemspec: Use Ruby to pick the files to keep

olleolleolle:
This also excludes the gemspec from the files to ship.

### DIFF
--- a/pagy.gemspec
+++ b/pagy.gemspec
@@ -11,6 +11,6 @@ Gem::Specification.new do |s|
   s.description   = 'Agnostic pagination in plain ruby: it works with any framework, ORM and DB type, with all kinds of collections, even pre-paginated, scopes, Arrays, JSON data... Easy, powerful, fast and light.'
   s.homepage      = 'https://github.com/ddnexus/pagy'
   s.license       = 'MIT'
-  s.files         = `git ls-files -z`.split("\x0").select{|f| f.start_with?('lib', 'pagy.gemspec', 'LICENSE') }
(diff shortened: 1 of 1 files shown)


ddnexus:
I prefer the code in this PR, but it will pull in also temporary files or excluded files that happen to be in the lib dir at the moment of the build. The git command would avoid that.

olleolleolle:
I see your point.

One way to get there would be a fully explicit file list, or to work with Rake clean/clobber to point out any "known temporary artifacts", and let a rake build task depend on such a cleanup. 

Or, one can use the super-explicit rake-manifest gem, which keeps a manifest file and generates this kind of content.

Alright, now I've typed the alternative ways that I know of.

Ah, yes, one more thing. A reason to do it: There's a Rubocop-packaging plugin, with rules about how to make the life of a Debian package maintainer easier - they do repackaging of gems, and any shelling out of Ruby makes their work harder. The plugin has documentation which tries to explain their perspective. 

Close or keep this PR, it's your repository. Looks cool!

ddnexus:
Thank you for all the suggestions and the motivation of getting this done. I like the `rake-manifest` gem! That would be the solution to all the problems, and should not get in the way of adding/renaming files. It will explicitly fail, unless you remembered to generate the new manifest, which is a good thing IMO.
 
If you have time to do it, I would be very happy to merge your PR!
Thank you!

ddnexus:
Implemented! Thank you for the suggestions.

olleolleolle:
@ddnexus Best-possible outcome!

PR #298, opened 2021-04-13
Feature/add pagy ar

A-S-Baqtyan:


ddnexus:
Thank you for the PR.

It looks like you must change the pluralization for `ar`. Right now you used the `:one_other`. Instead you should use `:arabic` which has to be added in the [p11n.rb file](https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb).

The arabic language uses the `:zero, :one, :two, :few, :many, :other` plurals (as you can see [here](https://github.com/svenfuchs/rails-i18n/tree/master/rails/pluralization)). That logic should be added to the [p11n.rb file](https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb), and you should add the same plural keys in your translated file. The `ar` pluralization is more similar to the `:east_slavic` pluralization (used in the `ru` locale, for example). 

Please, let me know if the [instructions](https://github.com/ddnexus/pagy/tree/master/lib/locales) are clear enough, or if you need more assistance.

Thanks

A-S-Baqtyan:
Okay I will check thanks.

ddnexus:
@A-S-Baqtyan I added the missing pluralization, tests and corrected your file.
Please, checkout the `add-arabic-locale`, complete the `ar.yml` file where it has missing values, and create a new PR.

Thank you

PR #299, opened 2021-04-15
Add clarification to metadata docs

benkoshy:
### Why this PR?

* I was looking into the docs, trying to make sense of metadata: it wasn't readily apparent  or clear to me (with the utmost respect). I went digging into the source code, to find out what it was all about. This PR was born out of my wish to make it clearer to readers, to perhaps obviate the need to look further.

### What changes have I made?

Three separate commits so you can use/discard as required:

* Minor readability to changes.
* Clarifications with an example as to what the method produces.

### Deepest thanks

My deepest thanks to you for all the work you've put in, over the years.

Feel free to discard this PR without compunction, if it is not useful. 

regards
Ben

### DIFF
--- a/docs/extras/metadata.md
+++ b/docs/extras/metadata.md
@@ -25,6 +25,13 @@ render json: { data: records,
                pagy: pagy_metadata(pagy) }
 ```
 
+...and when you do this, you will get a ruby hash, which when converted to a JSON object, will give you the following which you can display as per your requirements:
+
+```javascript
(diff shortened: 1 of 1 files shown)


ddnexus:
Hey @benkoshy Thank you for the PR and the nice words!

While I always appreciate your good intention, it looks like the clarification is more useful to you than to the general public this time. However, I added a small clarification in the Javascript page (trying to answer better to your Jitter question).

Thank you!

PR #302, opened 2021-04-20
Add missing arabic keys

benkoshy:
* Used Google translate to add in the missing keys. 
* Amended the commit that you originally created. hopefully that's ok?


ddnexus:
Thank you. I tried that way yesterday, but you have to be careful, and pick the right alternatives from the google translations. 
For example, I've found that the `two` could be probably the same as the `one` (but when they are looking very different, google pick a different word to translate "items").

BTW, since I've just released the new version, this can wait till the next one. There are quite a lot of planned changes, so there is time.

One thing though, I've force-rebased the branch, so there are a lot of commits in the way now :)

PR #309, opened 2021-05-10
Add: instructions on running cypress locally [ci skip]

benkoshy:
Hi Domizio,

Please review: whether I made a mistake etc.

As always, feel free to close without compunction if it is not a useful.

rgds
Ben


ddnexus:
Thank you Ben!
I replaced the previous section with what you added in this PR in a new commit (887c7831d6b5fa3f69989a54bdcb228bcfd8c8f1), plus a few minor adjustments).

PR #311, opened 2021-06-13
fix(pagy): deprecated_var is a class method

rainerborene:
Otherwise it will trigger an exception like: 
> ActionView::Template::Error: undefined method `deprecated_var' for #<Pagy:0x00007f613c196028>


ddnexus:
Interesting!
I am wondering why the [tests](https://github.com/ddnexus/pagy/actions) don't trigger it.

Ruby/rails versions?

rainerborene:
@ddnexus because there is no test that initializes `pagy` with `anchor` parameter :scream:

ddnexus:
oh well .... 🤦

PR #315, opened 2021-06-18
Fix documents for Searchkick extra

molfar:
Mastering PR technique with easy tasks and testing Vscode new remote repository extension at the same time


ddnexus:
Thank you!

PR #318, opened 2021-06-23
Fixes for Meilisearch extra

molfar:
This PR fixes response format from Meilisearch mock object to match original one.


ddnexus:
That's the problem with mock classes :)
Thank you.

PR #322, opened 2021-08-02
chore: add github action - publish & tag versions

benkoshy:
### What is this?

This commit allows for the automation of: (i) publishing ruby gems, and (ii) if the gem is successfully published, tagging the branch with pertinent version.

### How does it work?

 * The gem is bundled and pushed to ruby gems with a simple script. 
 * The tagging occurs based on the version contained in the pagy.js file. It is absolutely essential that both the pagy.rb  and pagy.js file have the same version numbers otherwise confusion might result.

### How do I run it?

 * First increment the versions in both the pagy.rb and pagy.js
    files.
 * Manually run the action from the actions tab in github.com.

image

 * You will also need to add github secrets, an API key to authenticate the publishing to ruby gems. Please call it `RUBYGEMS_AUTH_TOKEN`.

image

### Why?

  * To hopefully make things a little faster and easier.

### Why has a conditional been added to the pagy.js file?

 * We are using a node based script to run a 'step' which tags the default branch with the gem's version number. This revision    number is extracted from the pagy.js file. That file has a 'window' variable contained in it. But node has no concept of a 'window' variable. Granted, a dummy window variable can be added to the node script, but I opted for a conditional, out of convenience.

* We wanted to make the pagy.js file work for all browsers. It would not do for us to simply add a default export line to the pagy.js to extract the version number, because it would not be widely supported across all browsers. Hence, in this action we are reading the pagy.js file and then eval-ing it. Making the version number readily available to us.

### DIFF
--- a/.github/workflows/publish-gem-and-tag.yml
+++ b/.github/workflows/publish-gem-and-tag.yml
@@ -1,0 +1,56 @@
+
+# What does this action do?
+
+# This Github Action allows for: 
+#   (i) the publishing of the gem to ruby gems.
+#   (ii) automatic tagging of the default branch. 
+# The action is triggered through a manual action - click and run (i.e. through workflow_dispatch)
+# The default branch is assumed.
+
+# The version number of pagy is located in two distinct files: the lib/pagy.rb file and also the
+# lib/javascripts/pagy.js file. 
+
+# Ruby gems publishes the version number located in the pagy.rb file, while we tag the default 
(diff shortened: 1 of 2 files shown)


benkoshy:
(I also might want to add a sanity check test: to ensure that the versions on both the `pagy.rb` file and the `pagy.js` file are the same. 

I was thinking to output both in HTML write a cypress test etc. to make sure that they are both the same?

ddnexus:
There is already a test that would fail if the version in the js does not match with the rb file. Thanks

Issue #321, opened 2021-08-01
pagy*_nav_js methods do not provide a non-JavaScript version of the links


nomis:
# Code Issue
- [x] I researched through the [documentation](https://ddnexus.github.io/pagy/), the [pagy issues](https://github.com/ddnexus/pagy/issues) and a known Search Engine, and there is no reference/report/post that solves this problem
- [x] I did install/upgraded to the latest version of pagy (or the latest `3.*` for pagy legacy)

### Ruby/Rails/Sinatra/... Versions

* Ruby 2.5.1
* Rails 6.1.1
* Pagy 4.10.1

### Description

None of the `pagy*_nav_js` methods includes a non-JavaScript version of the links, so there is no fallback to HTML for web browsers without JavaScript support (or with JavaScript disabled). The "combo" version has working Back/Next links but the page selection doesn't work because it requires JavaScript.

I want to take advantage of the JavaScript navigation without losing basic HTML compatibility. The `nav` elements should contain normal links that are then replaced by JavaScript at runtime.

### Steps to reproduce the issue using the provided support

Using the standard `pagy_nav_js` function:
```
<%== pagy_nav_js(...) %>
```

The HTML output does not contain any content within the `nav` element, only data attributes for use by javaScript code:
```
[...]

PR #323, opened 2021-08-02
skip ci: docs - add note in case javascript is not enabled

benkoshy:
Addresses #321.


ddnexus:
Thank you @benkoshy. I integrated it with 71871016.

PR #324, opened 2021-08-07
Add: link directly to docs page

benkoshy:
### What / Why this PR?

Right now, if someone wants to update the docs page, they have to go to the repo and manually find the page.

When they click on the link, it would be nice if they were directed to the actual page which they are required to edit. This PR attempts to fix that problem.

image

**UPDATE:**

* I have built the site using `pagy-docker` and it looks good to me. Clicking the link will direct you to the relevant docs page where PRs against the master branch can be made in-line.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -16,7 +16,7 @@
 <body>
 <div class="wrapper">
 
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you

benkoshy:
I've put the wrong link. I will post a fix shortly.

Issue #307, opened 2021-05-04
summary sentence


m-abdurrehman:
Is there any way/support to have some sort of summary sentence? like

Showing 1 to 10 of 97 results
Showing 1 to 10 of 97 total
etc.

If not, I think its good to add this so user know the state with some stats

ddnexus:
You know, I spent a lot of time to write a very detailed doc, and it would be very nice if the users would at least try to search before just asking, don't you think?

https://ddnexus.github.io/pagy/api/frontend

benkoshy:
@m-abdurrehman thank you for posting the issue.  

* does this method do what you are after: https://ddnexus.github.io/pagy/api/frontend#pagy_infopagy-pagy_id--item_name--i18n_key- or please otherwise do shed some light on what you are trying to do.

* more importantly, if the above method does what you require, I would be very, very interested to hear your point of view on how things can be improved (e.g. whether documentation or otherwise).

chrs

Ben

m-abdurrehman:
@benkoshy thanks for sharing it, I missed it, it worked!

If you do not mind, please do share a link or page where I can play around with different styles of pagination controls in different front ends.

Just my feeling, documentation can be improved by using words commonly used(I mean words dictionary) and more examples

Thanks again @benkoshy , and also @ddnexus for your great work and time.

Another issue, I am having in my project is that only text is clickable(leads to next page when using with `tailwind`)
I used pagy in an other project with `bootstrap` and whole `li` was clickable.

Not sure if I am missing something or it is by design

DOM/UI when using **bootstrap**

![tailwind](https://user-images.githubusercontent.com/59817964/122194883-4a88a580-ceaf-11eb-8ed4-9f640b82d081.png)


DOM/UI when using **tailwind**

![bootstrap](https://user-images.githubusercontent.com/59817964/122194869-46f51e80-ceaf-11eb-8193-573e2540dede.png)

benkoshy:
Hi @m-abdurrehman 

### (a) different front ends with pagination controls: 

> If you do not mind, please do share a link or page where I can play around with different styles of pagination controls in different front ends.

Try using this repository here: https://github.com/benkoshy/pagy-rails

> documentation can be improved by using words commonly used(I mean words dictionary) and more examples

### (b) improvements suggested:

Can you please elaborate on your ideas here?

### (c) tailwind issue with links: i will check this out

> Another issue, I am having in my project is that only text is clickable(leads to next page when using with tailwind)
[...]

PR #333, opened 2021-09-09
simplify readme

benkoshy:
See here for a live preview: https://github.com/benkoshy/pagy/tree/refactor-readme

## What is this PR?

   * Sections put into collasible drop downs.
   * Summary image showing what you get with pagy.

## Why this PR?

* My concern is that the readme is too long, and that
* people might skip the readme, and ask questions that are perhaps unecessary. i.e. the 'less is more' philosophy.

With the approach suggested in this PR, users will read only what they want to read. Most new users will simply want to get started. More advanced users will drill down into the sections only when they need to. Most of it can be hidden and shown only when required.

* A summary image is shown so people can quickly see what you can get: to avoid issues like: #307.

Above all, I wanted to simplify and restyle to ease the burden on the maintainer, and make it slightly less cluttered and aesthetic for users.

These are simply suggestions/ideas - there are pros/cons with any decisions, and this PR should be discarded without compunction if deemded unprofitable.


ddnexus:
Thank you! There are a few improvements that I will definitely consider!

I love the collapsible summaries: that's a great idea... but I would never ever hide the charts from prominent view 😎 and just to put a bare pict of a pagination bar that everybody knows and every other pagination gem does. 😱 (not totally wrong, but just not the first thing I want people to see).

The thumbnails of the screen casts are huge so they must be hidden or better reduced and re-positioned maybe in 2 columns.

The "New in ..." section contains some general info (e.g. version) that should go at the top, however most of the remaining content should probably be hidden somewhere.

I will review your proposed changes in details and eventually include in a refactoring ASAP.

Thank you so much for your help!

benkoshy:
After reading your response, I fear that I have not communicated these ideas effectively – I shall rehash them perhaps a little better:

#### The purpose of the Github Readme – use it to fwd to the documentation site

* We now have at our disposal a Jekyl documentation site – which allows us to link / format / display better than a readme.md can probably ever hope to do. Consequently, I propose to move people off the readme and onto the documentation site as fast as possible – and leave the readme minimal. 

#### Favour Links over Text

* The volume of text might be overwhelming for new comers. I would favour breaking it all out into discrete pages that people can read if they want to. For example, I would suggest that the section on performance be entirely moved into a separate page (or onto the Detailed Gems Comparison page). I know that the most strenuous efforts were made to improve performance, and to produce charts, and that performance is the corner stone of the library – still, I would move it off entirely, with a link to it.

image

#### The Nature of the Questions being asked on issues / gitter

* I have been following issues / questions being asked for some time. I also have the added benefit of not writing the library – hence my familiarity with it, and all the features will not be as complete as the chief maintainer. Most of the questions suggest that folks are **unable to easily find the answers within the documentation**:
* See this question for example: https://github.com/ddnexus/pagy/issues/307#issue-875210198 – the question suggests a feature which is already existing. But for a new comer – this is perhaps not immediately obvious? 
* Here is another example: 

image

I propose changes to perhaps make it easier and more obvious so people can find what they are looking for, and not to overwhelm with text, in light of the questions being asked on issues / gitter channels. Only very slight changes may yield some huge gains for both the maintainer and users of the gem.

### Proposed Structure (with links)

What is pagy?
Why would I use it?
What can I get out of it? (Use Cases)      
            * API use cases.
            * Non api use cases (With pictures)
            * What features are available to me?
Tutorials
Changelogs
Support

Please let me know your thoughts.

ddnexus:
> The purpose of the Github Readme – use it to fwd to the documentation site

Not really.

The Readme should be a sort of pamphlet to help you to decide (in the shortest amount of time) whether the gem is worth your attention or not. 

On the other hand the documentation should contain "read-as-you-need" content that is useful when you already decided to **actually use** the gem. (and obviously confirm that is worth using it)

If you use the Readme just as an index pointing to the documentation you either are assuming that the users already decided to use it or that they are willing to research more and somewhere else (links) in order to decide if it's worth or not. Both assumption are usually quite far from reality and directly affect the most sensitive motivation for adoption. (There are also other facts related to this matter that I omit for brevity)

> leave the readme minimal

What belongs to the scope of the Readme explained above should remain into the Readme. If there are technical details not useful to showcase anything outstanding for adoption or useful to the overview of the gem, then they should go into the doc.

> Favour Links over Text

Not when the content shows something outstanding that we want the new user to know about pagy.

> I would move it off entirely, with a link to it.

Not going to happen. "An image is worth one thousand words" and 4 charts showing outstanding facts are worth 4 millions words. That charts must not even be hidden in a summary. In the same time I require the user to make a decision to click on the summary arrow, I can get his/her attention and communicate the most important and memorable reason s/he will choose pagy over anything else.

HTH

> folks are unable to easily find the answers within the documentation

yeah even if it is in the doc... even if it is just one google-search away...
Well... that's a fact even for very complete documentation. I guess users have something to do with it :)

> I propose changes to perhaps make it easier and more obvious so people can find what they are looking for

Great! What you are proposing is basically a Q&A section. 

I love the idea and you are probably the best guy for the job!

If you have time to dedicate to it, I would love your contribution organizing a list of basic questions for new comers that would have a prominent part in a nested summary right in the readme.

benkoshy:
Hey Domizio,

I think I understand the policy directive lucidly explained. Given that directive, the readme and documentation stands perfectly - I cannot fault it. 

> If you have time to dedicate to it, I would love your contribution organizing a list of basic questions for new comers that would have a prominent part in a nested summary right in the readme.

I can do this - but I do not have a clear understanding of the requirements?

(As an aside: perhaps my ideas might have come across as peremptory demands? I apologize if it came across like that - i mean them only as ideas / suggestions to be accepted or dropped according to their merit). Pls do let me know your ideas on FAQs etc.

rgds

Ben

ddnexus:
Ben, stop worrying! You are great! And too kind! :)

The Q&A section could be actually 2 sections. One with the basic questions useful to people that want to know what pagy offers. Stuff like:

What is pagy?
Why would I use it?
What can I get out of it? (Use Cases)
What are the common use cases (With pictures)
What are the API use cases?
What features are available to me?

 with answers that don't necessarily need to be short (remember, the user clicked to see the answer, so s/he wants to know right away without jumping around in another place). It should inform and end showcase the goodies :). No problem if we repeat content scattered around the doc.

A second section (if needed) could be more technical. We already have a long "How to", but we don't have something like "Troubleshooting" with common case of failures (e.g. "Pagy serves always the first page even if there are many") and that kind of stuff that may be common mistakes. You may find many of those in the Support Chat. Just a suggestion, but you are the Customer Service guy, so you know better than me what are the problems to address.

benkoshy:
Ok let me see what I can come up with by next Friday.

PR #335, opened 2021-09-23
Update docs: clarify vs code set up instructions

benkoshy:
Very minor changes to clarify questions users may have. 

These are suggestions, please consider and delete / accept / modify as required. rgds Ben


ddnexus:
Thank you Ben!

I changed:
- the link (is different)
- the formatting (it was inline)
- the "navigate to" suggestion (it was wrong)

I specified either "local terminal" or "container terminal" in the whole doc, since that is the probable cause of the confusion to navigate around.

Thank you!

PR #338, opened 2021-10-19
add: faq

benkoshy:
@ddnexus 

Domizio sir,

The following PR seeks to:
* Close out  #333.
* These are but suggestions - please feel free to amend / delete / close according to your discretion - if they are not beneficial etc., or LMK if further details / clarity is needed.

rgds

Ben


benkoshy:
> Hi Ben, while this info is certainly useful for very beginners, it looks unlikely that beginners would land to the pagy readme without already knowing what pagination is and why they need it, and finding it here for seniors is kind of lame.
> 
> I think you have a special gift to make things simple for beginner, so this should probably go in a more useful post that we can link prominently in the README, and without being relegated in a README section as it would be now.
> 
> What do you think?

My thoughts: 

- if you deem it to be of little benefit, then it should be removed - otherwise it will take up needless space for no benefit. I suppose the same reasoning could apply to the suggested post: will it be useful? Not sure, I might forbear, until evidence is presented to the contrary.

ddnexus:
Ben, I do think it is very useful information for very beginners, but the README is not a place where beginners will ever look for that info (i.e. if you are a beginner you have to know what pagination is before landing on the README), so in the README is not useful and worse, it is burying good info in the a place where nobody needs that will find it. That's why I think the info should be published because it is useful, and it should be in a mini-post somewhere, linked in the README for good measure.

benkoshy:
@ddnexus I think I understand: if it will be useful, I will prepare a short post on the topic.

benkoshy:
I have knocked-up a [basic beginner's post](https://benkoshy.github.io/2021/11/03/pagination-basics.html).  if interested, pls look it over and let me know your thoughts.

ddnexus:
Really nice post Ben!

PR #346, opened 2021-11-12
add: image showing sinatra app

benkoshy:
Why this PR?

   Makes it super easy to see the benefits with an image. speaks x1000 words


ddnexus:
I moved it a bit. Tank you Ben!

PR #347, opened 2021-11-19
Update calendar.md

benkoshy:
Why this PR?

* Easy cut and paste conveniences (to make it easier for developers).


ddnexus:
Thanks. I added one of the two lines, in a different part of the docs that has been reorganized. Still missing some example though.

PR #349, opened 2021-11-28
Add Tamil (ta) translation

jyuvaraj03:
Came across this gem when I was developing an application with Tamil language support. I've already added translation to the necessary (pagy) strings inside my application. Sharing the translations here!

The pluralization rule for Tamil is `OneOther`. But I did not find the pluralization rule file in [rails-i18n list](https://github.com/svenfuchs/rails-i18n/tree/master/rails/pluralization). If it's needed, I'd be happy to raise a PR there too to include the pluralization rule. 

P.S. If any fellow Tamil-speaker comes across this, feel free to correct me if the translations can be improved!

### DIFF
--- a/lib/locales/ta.yml
+++ b/lib/locales/ta.yml
@@ -1,0 +1,22 @@
+# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+
+ta:
+  pagy:
+
+    item_name:
+      one: "பதிவு"
+      other: "பதிவுகள்"
+
+    nav:
+      prev: "&lsaquo;&nbsp;முந்தையது"
+      next: "அடுத்தது&nbsp;&rsaquo;"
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you for the PR!
The missing rails pluralization looks weird and might be a bug, unless the `i18n` defaults to `OneOther` for missing files, in which case is just an inconsistency, since other locales with `OneOther` pluralization do have their related file in the list.

You may add an issue there, so we will find out. Please, mention this issue, so we can keep track of it.
Thank you!

jyuvaraj03:
`OneOther` is in fact the [default pluralization rule](https://github.com/svenfuchs/rails-i18n/blob/f6cc43ec8cfc08d0db7a06ca8299d7834c517a05/lib/rails_i18n/common_pluralizations/one_other.rb#L1). So, it's just an inconsistency. I've raised a PR in `rails-18n` to add the missing rule.

PR #354, opened 2022-01-18
Add Ruby 3.1 to CI

petergoldstein:
This PR adds Ruby 3.1 to CI.

Other than the relatively obvious change to the CI configuration there are two other changes:

1. Setting the '--disable-error_highlight' RUBYOPT so that we don't get error highlighting in error messages in test
2. Updating Psych to explicitly permit instances of the Time class when loading YAML in the mocks.


ddnexus:
Thank you!

PR #355, opened 2022-01-18
Fix for wrong calculation of local time

ddnexus:
**IT PROBABLY DOES NOT WORK AND MUST BE REIMPLEMENTED**

Sorry no time to try it properly, so I have to guess for now. The conversion to UTC to local must be done when creating a new Time object, in order to find e.g. the start of the month for that time zone. The implementation in this branch creates the start of the periods in UTC and then converts the labels in local, which means that the local time is indicating a start of the month for UTC which translates to a local start of the month that has an offset.

---
- All Time input, output and calculations are expressed in UTC time now (e.g. `:period`, `from` and `to`)
- The `label_for` method calls the `:in_time_zone` user-defined proc variable in order to correctly convert the labels to local time
- The old incorrect code is deprecated but still supported till 6.0 in order to avoid breaking apps that may work without problems

### TO DO
- Tests for new code (old tests pass)
- Update `apps/pagy_calendar_app.ru`
- Review doc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ If you upgrade from version `< 5.0.0` see the following:
 Still supported in version `5.x` but not supported from `6.0` on:
 
 - `pagy_massage_params` method: use the `:params` variable set to a lambda `Proc` that does the same (but per instance). See [How to customize the params](https://ddnexus.github.io/pagy/how-to#customize-the-params).
-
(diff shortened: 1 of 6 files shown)


ddnexus:
@petergoldstein for some reason the GH reviewers UI does not work. Please, let me know if the change works as expected. Thanks

petergoldstein:
@ddnexus Got caught up in some stuff today, but I should be able to look at this around 9p Pacific.

ddnexus:
@petergoldstein by that time I hope I will have completed a different approach, because this one does not work.

PR #358, opened 2022-01-19
Fix pagy calendar calculation of local time

ddnexus:
- The Pagy::Calendar::* classes require ActiveSupport
- The :period items can be Time or TimeWithZone objects (which work better with TZ)
- Deprecated use of week `:offset` in favour of `:first_wday`
- Updated tests
- Updated docs


ddnexus:
@petergoldstein this one looks a lot better than the other one. Please, let me know if it works in your test cases. Also: it would be great if you could add some test to check for the specific problem you had. Thanks.

petergoldstein:
@ddnexus Will give it a look tonight (around 9p) my time.

ddnexus:
@petergoldstein Here is a high level summary: 

- The refactoring should not break the old usage (i.e. it will not require any change to legacy code), while it should handle the time zone calculations properly
  - If you pass a plain Time object it will work as before
  - If you pass a TimeWithZone object it should fix the reported problems
  - The activesupport dependency has been added to the gemfile (to avoid to break the rare apps that don't already use it with Rails). In 6.0 it will be removed as a core dependency and replaced by a simple requirement note in the Calendar app doc.

petergoldstein:
@ddnexus Great.  Going to give it a first look tonight and dig in tomorrow.

ddnexus:
@petergoldstein The PR should be missing only a specific test covering your case to be good to release.

If you don't have the time to add it, please, just let me know the conditions to test. Thank you!

petergoldstein:
@ddnexus Basically the conditions you want to test (for the TimeWithZone) is:

1. For a time zone with Daylight Saving (e.g. Pacific) does it properly compute the end label when the range doesn't cross a Daylight Saving transition.  Test this for all interval types.
2. For a time zone with Daylight Saving (e.g. Pacific) does it properly compute the end label when the range crosses a Daylight Saving transition.  Test this for all interval types.
3. For a time zone without Daylight Saving (e.g GMT or Arizona) does it properly contribute the end label when the range doesn't cross a Daylight Saving transition.  Test this for all interval types.
4. For a time zone without Daylight Saving (e.g GMT or Arizona) does it properly contribute the end label when the range crosses a Daylight Saving transition.  Test this for all interval types.

ddnexus:
That's helpful thank you! 

Just to avoid any mistake interpreting your points: I should check the exact beginning time, since it can offset one hour, hence causing a possible day shift in the label, right?

PR #359, opened 2022-01-19
Add: from, offset, and to tests for gearbox testing

benkoshy:
* I may have missed an edge case. please let me know if i have. hope this helps.


ddnexus:
Thank you!

benkoshy:
the changes i made are poorly organised, and i missed also the offset for countless. i will push change shortly.

PR #364, opened 2022-01-31
Update README.md

benkoshy:
* fix typo, add link


ddnexus:
Typo fixed externally, link not added. Either we should add all the links to all the other tools referred in the page or simply ignore them. The user can easily find them in the regular documentation.
Thank you!

PR #365, opened 2022-02-01
fix: clarify javascript webpack instructions

benkoshy:
WHY?
* to clarify instructions, and minimize copy paste problems and support
requests.

* these are some questions that arose in my mind as i was following
these instructions.

as always, close / delete emend according to discretion.


ddnexus:
Thank you!

PR #367, opened 2022-02-02
Use ms_search for meilisearch to prevent conflicts

### LINKED ISSUE #366: Use explicit meilisearch method
I've got an app using Ransack and it provides a `search` method that conflicts with Meilisearch.

In the meilisearch extra, this calls `search` directly which in our case is the Ransack search method.
https://github.com/ddnexus/pagy/blob/8d37ace7b4b699c59e0509f5f9fd3cd42db63d83/lib/pagy/extras/meilisearch.rb#L38

Meilisearch provides an `ms_search` and aliases it as `search` to prevent conflicts:
https://github.com/meilisearch/meilisearch-rails/blob/main/lib/meilisearch-rails.rb#L308

Would you accept a PR changing the Meilisearch extra to directly use `ms_search` so that it avoids conflicts automatically?

excid3:
Fixes #366 by using the `ms_search` method directly from meilisearch-rails.

### DIFF
--- a/lib/pagy/extras/meilisearch.rb
+++ b/lib/pagy/extras/meilisearch.rb
@@ -35,7 +35,7 @@ class Pagy # :nodoc:
         vars                 = pagy_meilisearch_get_vars(nil, vars)
         options[:limit]      = vars[:items]
         options[:offset]     = (vars[:page] - 1) * vars[:items]
-        results              = model.search(term, **options)
+        results              = model.ms_search(term, **options)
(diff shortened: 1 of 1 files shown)


excid3:
Not entirely sure why tests are failing. Any pointers @ddnexus?

benkoshy:
@excid3 mocks are used hence the failing tests. I've issued another PR which fixes that. thx

PR #369, opened 2022-02-02
add: ms_search alias to prevent conflict method name conflict

benkoshy:
* fix:  #367 failing tests.

### DIFF
--- a/docs/extras/meilisearch.md
+++ b/docs/extras/meilisearch.md
@@ -20,7 +20,7 @@ require 'pagy/extras/meilisearch'
 If you have an already paginated `Meilisearch` results, you can get the `Pagy` object out of it:
 
 ```ruby
-@results = Model.search(nil, offset: 10, limit: 10, ...)
+@results = Model.ms_search(nil, offset: 10, limit: 10, ...)
 @pagy    = Pagy.new_from_meilisearch(@results, ...)
 ```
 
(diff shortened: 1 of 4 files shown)


benkoshy:
* I suppose that there might be conflicts with searchkick as well? because that also uses `Model.search`: but looks like they're recommending something like this: https://github.com/elastic/elasticsearch-rails/issues/96#issuecomment-49515465

for reference, [here's the ransack method](https://github.com/activerecord-hackery/ransack/blob/49a43fc9519408e47a4844d3cdbeb2838687041a/lib/ransack/adapters/active_record/base.rb#L7)

ddnexus:
@excid3 @benkoshy It looks like ransack is smart enough to [skip the `search` alias if the model respond_to? it](https://github.com/activerecord-hackery/ransack/blob/49a43fc9519408e47a4844d3cdbeb2838687041a/lib/ransack/adapters/active_record/base.rb#L7), so in theory if you load another gem defining `search` before loading ransack, and use `Model.ransack` instead of `Model.search`, we wouldn't even need to change anything.

However the `ms_search` would make things simpler for free, at least for `meilisearch`. For `searchkick`/`ransack`  and `elasticsearch-rails`/`ransack` the load order should make them work.  Not sure about `searchkick`/`elasticsearch-rails` but that combination looks quite unlikely.

I am more oriented toward a totally flexible solution: adding an extra variable in the `meilisearch`, `searchkick`, `elasticsearch-rails`, so the user could override the method that the extra calls in the model when needed.

Working on it.

excid3:
Our Ransack usage comes from ActiveAdmin, so I'm not sure when it loads and can't really control it.

It would still be better if this uses `ms_search` by default for Meilisearch. That way there's no load order to worry about.

ddnexus:
> It would still be better if this uses ms_search by default for Meilisearch. That way there's no load order to worry about and we don't have to add a configuration line.

With the current pagy version there is no loading problem, however you are right: the `ms_search` default is better, since it's the original method that is always defined. Earlier I thought that it was the `search` method.

ddnexus:
Changed in `5.10.1`

excid3:
Thanks @ddnexus, you're the best!

PR #372, opened 2022-02-04
Fix: docs - typo, minimise ambiguity

benkoshy:
WHY?

 * Encourage users to use `Model.ms_search` instead of `Model.search` to
avoid conflicting method name problems, due to incorrectly loading
order of Ransack, minimize ambuguity, and support requests


benkoshy:
sorry i accidentally added some test/e2e files before force pushing.

ddnexus:
Thank you!

PR #373, opened 2022-02-05
This setting for enabling fancy-routes works for my rails 7 project

charlietag:
* Ruby
  * 3.1.0
* Rails
  * 7.0.1
* Pagy
  * 5.10.1

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -293,8 +293,8 @@ The following is a Rails-specific alternative that supports fancy-routes (e.g. `
 
 ```ruby
 def pagy_url_for(pagy, page, absolute: false, html_escaped: false)  # it was (page, pagy) in previous versions
-  params = request.query_parameters.merge(pagy.vars[:page_param] => page )
-  url_for(params)
+  params = request.query_parameters.merge(pagy.vars[:page_param] => page, only_path: ! absolute )
+  html_escaped ? url_for(params).gsub('&', '&amp;') : url_for(params)
 end
 ```
 
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -292,7 +292,7 @@ The following are a couple of examples.
 The following is a Rails-specific alternative that supports fancy-routes (e.g. `get 'your_route(/:page)' ...` that produce paths like `your_route/23` instead of `your_route?page=23`):
 
 ```ruby
-def pagy_url_for(pagy, page)  # it was (page, pagy) in previous versions
+def pagy_url_for(pagy, page, absolute: false, html_escaped: false)  # it was (page, pagy) in previous versions
   params = request.query_parameters.merge(pagy.vars[:page_param] => page )
   url_for(params)
 end


ddnexus:
It should work also without the explicit params, since that is the default.

Am I missing something?

charlietag:
> It should work also without the explicit params, since that is the default, which BTW the method is not using.
> 
> Am I missing something?

I will get this error, if I do not define the explicit params

image

Or I missed something ?

charlietag:
Is it caused by this ?

https://github.com/ddnexus/pagy/blame/521a4c2057b81abd4e5eba78fcc18d2a5c046501/lib/pagy/extras/support.rb#L45-L50


image

ddnexus:
Oh, right. I forgot that in the doc it doesn't look like just an example. Then the following would probably be a more complete solution:

```ruby
def pagy_url_for(pagy, page, absolute: false, html_escaped: false)  # it was (page, pagy) in previous versions
  params = request.query_parameters.merge(pagy.vars[:page_param] => page, only_path: ! absolute )
  html_escaped ? url_for(params).gsub('&', '&amp;') : url_for(params)
end
```
I cannot try it right now. Please, could you confirm that it works?

charlietag:
> Oh, right. I forgot that in the doc it doesn't look like just an example. Then the following would probably be a more complete solution:
> 
> ```ruby
> def pagy_url_for(pagy, page, absolute: false, html_escaped: false)  # it was (page, pagy) in previous versions
>   params = request.query_parameters.merge(pagy.vars[:page_param] => page, only_path: ! absolute )
>   html_escaped ? url_for(params).gsub('&', '&amp;') : url_for(params)
> end
> ```
> 
> I cannot try it right now. Please, could you confirm that it works?

This also works for mine project !!

ddnexus:
Sorry, I didn't see your last commit. 
Thank you for your PR!

charlietag:
> Sorry, I didn't see your last commit. Thank you for your PR!

Thank you for the great pagination tool !

sedubois:
OK, it looks like the below code works. It will result in links such as `/my-slug/p/2` (I added the `/p` to avoid conflict with other sub-routes).

```rb
# config/routes.rb
resources :contents, path: '', only: %i(), constraints: proc { |req| ... } do
  get '(p/:page)', on: :member, action: :show, as: ''
  # other nested routes, e.g.
  get :about, on: :member
end

# app/controllers/contents_controller.rb
class ContentsController < ApplicationController
  include Pagy::Backend

  def show
    @content = Content.find_by_slug(params[:id])
    @pagy, @children = pagy(@content.children, items: 30)
  end
end

# app/helpers/application_helper.rb
module ApplicationHelper
  include Pagy::Frontend

  def pagy_url_for(pagy, page, absolute: false, html_escaped: false)
    url_for(page:)
  end

  # for some reason this needs to be redefined, as the Pagy::TrimExtra doesn't get picked up
  def pagy_link_proc(pagy, link_extra: '')
    link_proc = super(pagy, link_extra: link_extra)
    lambda do |page, text = pagy.label_for(page), extra = ''|
      link = +link_proc.call(page, text, extra)
      return link unless page == 1
      pagy_trim(pagy, link)
    end
  end

  # redefine the trimming for page 1
  def pagy_trim(pagy, link)
    link.sub!(/\/p\/1\b/, '')
  end
end
```

```erb
# app/views/contents/show.html.erb
<%== pagy_nav(@pagy) %>
```

PR #377, opened 2022-02-16
Add ability to add custom HTML classes for ul content

jtapia:
Trying to implement it over my app, I had the necessity to send/add some custom classes to the `<ul>` element, let me know what do you think about it

### DIFF
--- a/docs/api/frontend.md
+++ b/docs/api/frontend.md
@@ -44,6 +44,7 @@ This method takes the Pagy object and returns the HTML string with the paginatio
 The method accepts also a few optional keyword arguments:
 
 - `:pagy_id` which adds the `id` HTML attribute to the `nav` tag
+- `:pagy_classes` which use the passed Array to add the `class` HTML attribute to the `ul` tag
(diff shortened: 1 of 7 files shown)


ddnexus:
Thank you for the PR. Pretty well done!

It can be useful to others with your very specific need, however it covers only the `pagy_nav`  and `pagy_bootstrap_nav` helpers... no other frontend extras and no `pagy*_nav_js` helpers.

A principle that I tried to respect in pagy so far is that there should be no "asymmetries" between `*nav` and `*nav_js` nor between frontend extras... and since different frontends use even different HTML elements, the params that you can pass to the helpers should also be frontend-agnostic. 

That practically means that there are no params for that. Adding one would require a lot of gimmick and work because it would have to fit with all the rest, not to mention that it would introduce a bit of overhead also for all the apps that don't need it.

That's the reason of the doc in [How to - Customize CSS Styles](https://ddnexus.github.io/pagy/how-to#customize-css-styles).

Would that work in your app?

jtapia:
@ddnexus got it, thanks, will try with that

Issue #378, opened 2022-03-07
Includes and joins ignored with meilisearch extra


coder2000:
# Code Issue
- [x] I researched through the [documentation](https://ddnexus.github.io/pagy/), the [pagy issues](https://github.com/ddnexus/pagy/issues) and a known Search Engine, and there is no reference|report|post|video that solves this problem
- [x] I did install|upgraded to the latest version of pagy (or the latest `3.*` for pagy legacy)
- [x] I am providing at least one of the following working and self-contained code support that can reproduce this issue:
   
   - [ ] Simple step by step list that would work in IRB with the [Pagy::Console](https://ddnexus.github.io/pagy/api/console)
   - [ ] Plain ruby file that can run as `ruby my-problem.rb`
   - [ ] Edited copy of the single file [pagy_standalone_app.ru](https://github.com/ddnexus/pagy/blob/master/apps/pagy_standalone_app.ru)
   - [ ] Edited copy of the single file [pagy_bare_rails.rb](https://github.com/ddnexus/pagy/blob/master/apps/pagy_bare_rails.rb)
   - [ ] Link of my own branch forked from pagy, which contains an added test file  
   - [ ] Link of my own branch forked from one branch of the [Pagy Rails App](https://github.com/benkoshy/pagy-rails)
   - [ ] Link of a standalone `docker image` downloadable from a docker repository (e.g. dockerhub.com)
   - [ ] `docker-compose` file that can be run with `docker-compose up` (and doesn't need to be built because it uses only docker images and no local context)

### Ruby|Rails|Sinatra|... Versions
Ruby 3.1.0
Rails 7.0.2

### Description
Using `includes` with meilisearch extra doesn't execute the same queries and results in N+1.

### Steps to reproduce the issue using the provided support
Use `Model.includes` with `pagy_search`

Thank you!

benkoshy:
Here is a very incomplete ruby script which might serve as a starting point to further investigate the issue: I beg the patience of the pagy community because it is not complete - unfortunately i am constrained. but hopefully it will save time for anyone researching this issue:

1. Ensure sql lite is installed.
2. `docker pull getmeili/meilisearch:latest # Fetch the latest version of Meilisearch image from Docker Hub`
4. `docker run -it --rm -p 7700:7700 getmeili/meilisearch:latest ./meilisearch -aster-key=password`
5. run the ruby script: `ruby pagy_bare_rails.rb` in a separate daemon

### Notes:
[...]

PR #357, opened 2022-01-19
update: to test pagination with real data rows

benkoshy:
Please refer to #356 In order to reproduce it, some additional code was required - e.g. setting up connections, adding sqllite, with examples to allow for the easy reproduction of issues.

Hopefully, with this PR, we will see the perfect reproduction of issues.

I hope this helps.

PR #379, opened 2022-03-10
update: pagy_bare_rails script: for db testing

benkoshy:
### Why this PR?

* The use case seems to have developed from (https://github.com/ddnexus/pagy/pull/357)
and again in (https://github.com/ddnexus/pagy/issues/378). There seems to be a need to test with db, though it requires a dependency.

### What's changed?

* Add example for use with external services e.g. meilisearch. The
implementation might be better done with docker-compose? That use case
may present itself in future, however I am not able to do it at this second.

* Minor cleanup

### DIFF
--- a/apps/pagy_bare_rails.rb
+++ b/apps/pagy_bare_rails.rb
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
-# Self-contained, bare rails app, usable to play with pagy
-# and/or easily reproduce any pagy issue.
+# Basic Rails app to: (i) reproduce errors, (ii) experiment pagy.
 
-# Edit this file as you need
-
-# USAGE:
+## Setup
+# 1. Install SqLite3 (an in-memory DB) e.g.
+##   sudo apt install sqlite3
+# 2. Optional: If using external services e.g. Meilisearch
+##    docker pull getmeili/meilisearch:latest # Fetch latest Meilisearch image from Docker Hub
+##    docker run -it --rm -p 7700:7700 getmeili/meilisearch:latest ./meilisearch -master-key=password
+# 4. Run script in separate terminal window:
 #     ruby pagy_bare_rails.rb
 
 # NOTICE: if you get any installation error(s) with the following setup
@@ -18,8 +21,7 @@ gemfile(true) do
   source 'https://rubygems.org'
 
   # Debuggers
-  # gem 'debug'
-
+  gem 'debug'
   # gem 'debase'
   # gem 'ruby-debug-ide'
 
@@ -30,16 +32,18 @@ gemfile(true) do
   gem 'rails', '~> 6.1'
   gem 'actionpack'
   gem 'railties'
+  gem 'sqlite3'
+
+  ## Optional: Meilisearch example
+  # gem 'meilisearch-rails'
 end
 
-# Set up debugging tools
-# require "debug" ## etc
+require 'debug'
 
-# pagy initializer - add your requirements here:
+## Pagy Set-up
 # https://ddnexus.github.io/pagy/extras
 # https://ddnexus.github.io/pagy/api/backend
-
-# Edit this section as you need
+require 'pagy/extras/meilisearch'
 require 'pagy/extras/metadata'
 require 'pagy/extras/overflow'
 require 'pagy/extras/trim'
@@ -52,7 +56,6 @@ Pagy::DEFAULT[:overflow] = :empty_page
 # Rails set up:
 require 'action_controller/railtie'
 require 'active_record'
-
 require 'minitest/autorun' # runs tests automatically
 
 class TestApp < Rails::Application # :nodoc:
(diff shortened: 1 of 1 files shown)


apps/pagy_bare_rails.rb
```diff
@@ -1,11 +1,14 @@
...
-
-# Edit this file as you need
-
-# USAGE:
+# Basic Rails app to: (i) reproduce errors, (ii) experiment pagy.
+
+## Setup
+# 1. Install SqLite3 (an in-memory DB) e.g.
+##   sudo apt install sqlite3
+# 2. Optional: If using external services e.g. Meiliesearch
```
brunoocasali, line 8:
```suggestion
# 2. Optional: If using external services e.g. Meilisearch
```

apps/pagy_bare_rails.rb
```diff
@@ -68,18 +71,72 @@ class TestApp < Rails::Application # :nodoc:
   routes.draw do
     get '/' => 'test#index', as: 'test'
   end
+
+  ## Optional: MeiliSeach example
```
brunoocasali, line 75:
```suggestion
  ## Optional: Meiliseach example
```

apps/pagy_bare_rails.rb
```diff
@@ -68,18 +71,72 @@ class TestApp < Rails::Application # :nodoc:
...
+  t.text :name
+  t.references :author, foreign_key: true
+end
+
+ActiveRecord::Base.logger = Logger.new($stdout)
+
+class Author < ActiveRecord::Base
+  has_many :books
+
+  ## Optional: MeiliSeach example
```
brunoocasali, line 96:
```suggestion
  ## Optional: Meiliseach example
```

apps/pagy_bare_rails.rb
```diff
@@ -68,18 +71,72 @@ class TestApp < Rails::Application # :nodoc:
...
+  #
+  # meilisearch do
+  #  attribute :name
+  # end
+end # :nodoc:
+
+class Book < ActiveRecord::Base
+  belongs_to :author
+
+  ## Optional: MeiliSeach example
```
brunoocasali, line 108:
```suggestion
  ## Optional: Meiliseach example
```

apps/pagy_bare_rails.rb
```diff
@@ -68,18 +71,72 @@ class TestApp < Rails::Application # :nodoc:
...
+
+1.upto(11) do |i|
+  Author.create(name: i)
+end
+
+Author.all.each_with_index do |author, i|
+  Book.create(author: author, name: i)
+end
+
+## Optional: MeiliSeach example
```
brunoocasali, line 124:
```suggestion
## Optional: Meiliseach example
```

apps/pagy_bare_rails.rb
```diff
@@ -68,18 +71,72 @@ class TestApp < Rails::Application # :nodoc:
...
   include Rails.application.routes.url_helpers
-
   include Pagy::Backend
 
   def index
-    meta, _records = pagy(Record.none, items: 10)
-    render json: { data: [], meta: pagy_metadata(meta) }
+    meta, @books = pagy(Book.all, items: 10)
+
+    ## Optional: MeiliSeach example
```
brunoocasali, line 134:
```suggestion
    ## Optional: Meiliseach example
```

brunoocasali (commented):
Hey @benkoshy just sending my 2cents here, I'm not the maintainer of this project, but I made some suggestions regarding the @meilisearch brand :)

Besides my own interest in these changes (I work at meili), I'm not sure if they should be here in this `pagy_bare_rails.rb` (since it is "bare rails"), maybe you could add them to a new file like this bare_rails one?

Thanks anyway for using Meilisearch 🤘

apps/pagy_bare_rails.rb
benkoshy: agreed. i"ll fix the spelling

apps/pagy_bare_rails.rb
benkoshy: agreed: Meilisearch - with sea**R**ch vs. seach - i'll make this change unless otherwise noted.

apps/pagy_bare_rails.rb
benkoshy: will modify as above

apps/pagy_bare_rails.rb
benkoshy: will modify as above

apps/pagy_bare_rails.rb
benkoshy: will modify as above

apps/pagy_bare_rails.rb
benkoshy: will modify as above

apps/pagy_bare_rails.rb
brunoocasali: Oh my, I didn't notice that! good catch! :p

benkoshy:
[brunoocasali](https://github.com/brunoocasali) I will make the changes you've suggested.

Your point of having a separate script for Meilisearch is noted, and warrants some thought. i thank you for suggesting it.

Issue #381, opened 2022-03-29
`pagy_bootstrap_combo_nav_js` page selector is rendered as disabled button


adas172002:
# Code Issue
- [x] I researched through the [documentation](https://ddnexus.github.io/pagy/), the [pagy issues](https://github.com/ddnexus/pagy/issues) and a known Search Engine, and there is no reference|report|post|video that solves this problem
- [x] I did install|upgraded to the latest version of pagy (or the latest `3.*` for pagy legacy)
- [ ] I am providing at least one of the following working and self-contained code support that can reproduce this issue:
   
   - [ ] Simple step by step list that would work in IRB with the [Pagy::Console](https://ddnexus.github.io/pagy/api/console)
   - [ ] Plain ruby file that can run as `ruby my-problem.rb`
   - [ ] Edited copy of the single file [pagy_standalone_app.ru](https://github.com/ddnexus/pagy/blob/master/apps/pagy_standalone_app.ru)
   - [ ] Edited copy of the single file [pagy_bare_rails.rb](https://github.com/ddnexus/pagy/blob/master/apps/pagy_bare_rails.rb)
   - [ ] Link of my own branch forked from pagy, which contains an added test file  
   - [ ] Link of my own branch forked from one branch of the [Pagy Rails App](https://github.com/benkoshy/pagy-rails)
   - [ ] Link of a standalone `docker image` downloadable from a docker repository (e.g. dockerhub.com)
   - [ ] `docker-compose` file that can be run with `docker-compose up` (and doesn't need to be built because it uses only docker images and no local context)

### Ruby|Rails|Sinatra|… Versions
ruby 2.7.3, ror 7.0.2.3 with `importmap-rails` gem, bootstrap 5.1.3 installed via official `bootstrap` gem w/o jQuery
...

### Description
`pagy_bootstrap_combo_nav_js` is rendering the following HTML code, with page selector input disabled

```
<div class="pagy-combo-input btn btn-primary disabled" style="white-space: nowrap;">
   <label>Page <input type="number" min="1" max="26" value="6" class="text-primary" style="padding: 0; border: none; text-align: center; width: 3rem;"> of 26</label>
</div>
```

### Steps to reproduce the issue using the provided support
I found no similar issues here.

benkoshy:
Right. Apologies,  I had forgotten about it. 

I'll post a fix shortly.

benkoshy:
[...]

PR #388, opened 2022-05-12
Fix bootstrap combo nav js

benkoshy:
### What is this?
Fix for #381

### What was done?

* Previously when using the combo box, you weren't able to select the input and type in the page you want to go to.

Why not?

* We had a parent element which had the 'disabled' class applied to it. This rendered descended child elements also disabled. But we don't want them disabled: we want to allow people to type in page numbers when they want to. Using in-line styles, we have basically overriden the 'disabled' input by allow pointer access to apply, as per normal, to the input element.
* I have not checked for WAI-ARIA compliance. It probably isn't complaint.

(This solution doesn't touch the styling of bootstrap, though I have checked three main versions of bootstrap against this solution. And it seems to work as before.)


benkoshy:
Sorry I'll cherry pick against master.

PR #389, opened 2022-05-12
fix ui bug - #381: input was disabled

benkoshy:
### What is this?
Fix for https://github.com/ddnexus/pagy/issues/381

### What was done?
Previously when using the combo box, you weren't able to select the input and type in the page you want to go to.

#### Why not?

- We had a parent element which had the 'disabled' class applied to it. This rendered descended child elements also disabled. But we don't want them disabled: we want to allow people to type in page numbers when they want to. Using in-line styles, we have basically overriden the 'disabled' input by allow pointer access to apply, as per normal, to the input element.
- I have not checked for WAI-ARIA compliance. It probably isn't complaint.

(This solution doesn't touch the styling of bootstrap, though I have checked three main versions of bootstrap against this solution. And it seems to work as before.)

Hope this helps.

### DIFF
--- a/lib/pagy/extras/bootstrap.rb
+++ b/lib/pagy/extras/bootstrap.rb
@@ -51,8 +51,8 @@ class Pagy # :nodoc:
       p_page  = pagy.page
       p_pages = pagy.pages
       input   = %(<input type="number" min="1" max="#{p_pages}" value="#{
-                    p_page}" class="text-primary" style="padding: 0; border: none; text-align: center; width: #{
-                    p_pages.to_s.length + 1}rem;">)
+                    p_page}" class="text-primary" style="pointer-events: auto; padding: 0; border: none; text-align: center;
+										width: #{p_pages.to_s.length + 1}rem;">)
 
       %(<nav#{p_id} class="pagy-bootstrap-combo-nav-js pagination" aria-label="pager"><div class="btn-group" role="group" #{
           pagy_data(pagy, :combo, pagy_marked_link(link))}>#{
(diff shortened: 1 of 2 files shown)


benkoshy:
Not sure why RuboCop is failing. I"ll have to come back to this. There is also the E2E that's failing. So this is a WIP in progress.

ddnexus:
Ben, Thank you for your work!

If you have problems with the custom CSS in the element, we might simply use a different native bootstrap style (if there is one that suits our needs). The goal is giving the central part a different color than the buttons, so any alternative that achieve that would be fine.

ddnexus:
About the failing rubocop, you may want to rebase on the `dev` branch which includes the changes needed for the new updated rubocop version.

benkoshy:
> If you have problems with the custom CSS in the element, we might simply use a different native bootstrap style (if there is one that suits our needs). The goal is giving the central part a different color than the buttons, so any alternative that achieve that would be fine.

I did try this - for example changing the background color, but found that bootstrap version 3 had some unexpected results where the UI was unacceptably changed. I shall try again, rebasing, and putting: 

`background-color: #8064A2 !important;` inline, and I will report back on how that goes with a new PR.

ddnexus:
We cannot use an hard-coded color, which may not play well with the rest of the app where it is used. We can only use bootstrap native classes that are consistently used in its themes, so when you change the theme the Pagy combo will adapt to it, OR use some color-agnostic trick that will be applied regardless of the class.

benkoshy:
> We cannot use an hard-coded color, which may not play well with the rest of the app where it is used. We can only use bootstrap native classes that are consistently used in its themes, so when you change the theme the Pagy combo will adapt to it, OR use some color-agnostic trick that will be applied regardless of the class.

Good point. Agreed.  I will come at it from a different angle.

PR #390, opened 2022-06-10
Fix NoMethodError when page param is a hash

yenshirak:
`NoMethodError` is raised when the `page` param is a hash, e.g., `page[something]=1`. This can be caused by bots that are trying to find vulnerabilities by manipulating the params.

### Steps to replicate
Run the self-contained Sinatra/Rails app.
Set the `page` param to a hash, e.g., http://localhost:8080/?page[something]=1

#### Expected
`Pagy::VariableError` is raised.

#### Actual
`NoMethodError` is raised: `undefined method 'to_i' for {"something"=>"1"}`.

### DIFF
--- a/lib/pagy.rb
+++ b/lib/pagy.rb
@@ -92,7 +92,7 @@ class Pagy
   def setup_vars(name_min)
     name_min.each do |name, min|
       raise VariableError.new(self, name, ">= #{min}", @vars[name]) \
-            unless @vars[name] && instance_variable_set(:"@#{name}", @vars[name].to_i) >= min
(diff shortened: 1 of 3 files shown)


ddnexus:
It makes sense. Thank you!

yenshirak:
@ddnexus I can fix the RuboCop issues as well unless you would like to disable these new cops.

ddnexus:
I think I already fixed that in `dev` and in `v6`. Rebasing your commit on top of `v6` should solve the issue.
Thanks

yenshirak:
@ddnexus done

ddnexus:
Thank you!

PR #392, opened 2022-06-22
Add Dependabot for GitHub Actions

petergoldstein:
As the checkout action in the configuration is out of date, Dependabot should open a PR to update the action after this is merged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,6 +50,11 @@ Style/CommentedKeyword:
 Style/EmptyCaseCondition:
   Enabled: false
 
+# cannot inline the disabling in the file because rubocop complains anyway
+Style/RedundantInitialize:
+  Exclude:
+    - test/mock_helpers/arel.rb
+
 # these cops are disabled only in the test files in order to
 # allow to copy and paste the failed output for test reconciliation
 Style/StringLiterals:
--- a/test/mock_helpers/arel.rb
+++ b/test/mock_helpers/arel.rb
@@ -15,9 +15,7 @@ module Arel
 
   module Nodes
     class Grouping
-      def initialize(_)
-        puts ""     # rubocop bug!
-      end
+      def initialize(_); end
     end
   end
 end


.rubocop.yml
```diff
@@ -50,6 +50,10 @@ Style/CommentedKeyword:
 Style/EmptyCaseCondition:
   Enabled: false
 
+Style/RedundantInitialize:
+  Exclude:
+    - test/mock_helpers/arel.rb
+
```
ddnexus, line 56: This solves the bug in rubocop that complains for the "redundant initializer" and also complains for the `Unnecessary disabling of Style/RedundantInitialize` if you disable it with inlining `# rubocop:disable Style/RedundantInitialize` 🤷

In the `v6` branch I added a [dummy line](https://github.com/ddnexus/pagy/blob/0cc6b06ecc4951568ffb4a5b50bd342d21b1a5a3/test/mock_helpers/arel.rb#L19) in order to avoid the global exclude of a single file, wondering which one is "better"

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
ddnexus, line 6: hmmm... I actually completely disabled dependabot because it complains for too many things that are not important for the gem itself. Just too annoying to keep up with it 🙃

ddnexus: Just noticed (thanks @benkoshy) that is checking the github-actions, which is actually very useful!

.rubocop.yml
petergoldstein: @ddnexus Just let me know how you'd prefer to handle it and I can make any change necessary.  I don't have a strong preference here.

ddnexus: I'll go with yours. 

Please, could you revert my dummy line, and add some comment like `# cannot inline the disabling in the file because rubocop complains anyway` comment on top of the exclude in `.rubocop.yml`. Thank you. 

_(I tried to do it myself rebasing your branch on `v6` but when I tried to push it to your branch, there was a long list of commits added below yours, and I didn't wanted to mess up your branch. Github `Squash and merge` will sort it out.)_

petergoldstein: @ddnexus I've made these changes.  Please let me know if anything else is required.  Thanks.

ddnexus:
Thank you @petergoldstein

Issue #394, opened 2022-07-14
Meilisearch 0.28.0 / meilisearch-rails 0.7.0 API breaking changes


benbws:
# Code Issue
- [x] I researched through the [documentation](https://ddnexus.github.io/pagy/), the [pagy issues](https://github.com/ddnexus/pagy/issues) and a known Search Engine, and there is no reference|report|post|video that solves this problem
- [x] I did install|upgraded to the latest version of pagy (or the latest `3.*` for pagy legacy)
- [x] I am providing at least one of the following working and self-contained code support that can reproduce this issue:
   - [x] Edited copy of the single file [pagy_bare_rails.rb](https://github.com/ddnexus/pagy/blob/master/apps/pagy_bare_rails.rb)

### Ruby|Rails|Sinatra|... Versions

Tested on '~> 6.1' but will no doubt apply to all versions as this looks to be a specific pagy meilisearch extension issue.

### Description

[Meilisearch 0.28.0](https://github.com/meilisearch/meilisearch) / [meilisearch-rails 0.7.0](https://github.com/meilisearch/meilisearch-rails) have introduced breaking API changes to results.raw_answer['nbHits'] used in https://github.com/ddnexus/pagy/blob/master/lib/pagy/extras/meilisearch.rb line 32 & 48

The immediate fix seems to be to replace 'nbHits' with 'estimatedTotalHits' but the Meilisearch documentation does recommend against using this method in general for pagination.

### Steps to reproduce the issue using the provided support

Meilisearch will need to be [installed](https://docs.meilisearch.com/learn/getting_started/quick_start.html#local-installation)/running locally on http://localhost:7700
ruby pagy_bare_rails.rb

Test will fail capturing the Pagy::VariableError - ERROR: expected :count >= 0; got nil

If you wish to remove the 'Record' index created in your local Meilisearch instance then run `curl -X DELETE 'http://localhost:7700/indexes/Record'`

### Edited pagy_bare_rails.rb
```
# frozen_string_literal: true

# Self-contained, bare rails app, usable to play with pagy
# and/or easily reproduce any pagy issue.

# Edit this file as you need

# USAGE:
#     ruby pagy_bare_rails.rb

# NOTICE: if you get any installation error(s) with the following setup
# temporarily remove `pagy/Gemfile` and `pagy/Gemfile.lock` from the repository
# because they may interfere with the bundler/inline.

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

[...]

PR #395, opened 2022-07-14
Fix for #394 - updated pagy/lib/pagy/extras/meilisearch.rb

benbws:
### What is this?
Fix for https://github.com/ddnexus/pagy/issues/394

### What was done?
nbHits renamed to estimatedTotalHits

[Meilisearch 0.28.0](https://github.com/meilisearch/meilisearch/releases/tag/v0.28.0) has introduced breaking changes: Changes in /indexes/{uid}/search endpoint - nbHits is renamed estimatedTotalHits. Some users were confused by the old name and used it for their pagination, which we discourage. Please check out this [fresh new guide](https://docs.meilisearch.com/learn/advanced/pagination.html) to learn how to paginate with Meilisearch without using nbHits.

### DIFF
--- a/lib/pagy/extras/meilisearch.rb
+++ b/lib/pagy/extras/meilisearch.rb
@@ -29,7 +29,7 @@ class Pagy # :nodoc:
       def new_from_meilisearch(results, vars = {})
         vars[:items] = results.raw_answer['limit']
         vars[:page]  = (results.raw_answer['offset'] / vars[:items]) + 1
-        vars[:count] = results.raw_answer['estimatedTotalHits']
+        vars[:count] = results.raw_answer.dig('estimatedTotalHits') || results.raw_answer.dig('nbHits')
         new(vars)
       end
     end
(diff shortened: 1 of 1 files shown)


brunoocasali:
Thanks for taking care of that, @benbws. I'm planning to introduce a class system to these returns to control these kinds of breaking changes in the following versions.

ddnexus:
@benbws Thank you for the PR.

It looks like changes would break backward compatibility. 

Something like the following would provide backward compatibility:

```rb
  key_name     = raw_answer.key?('estimatedTotalHits') ? 'estimatedTotalHits' : 'nbHits'
  vars[:count] = results.raw_answer[key_name]
```

Besides we should refactoring the tests in order to cover both cases.

ddnexus:
@brunoocasali Please, could you give me some example about what 'estimateTotalHits' is useful for? Also, what is the margin of error with the real count? Thanks.

ddnexus:
@benbws @brunoocasali See specific implementation discussion #397

benbws:
@ddnexus I've amended the lines to `vars[:count] = results.raw_answer.dig('estimatedTotalHits') || results.raw_answer.dig('nbHits')`. I went with this 'one liner' approach as I assume we'll revisit this later as per the discussion thread (which I will add to now).

brunoocasali:
> @brunoocasali Please, could you give me some example about what 'estimateTotalHits' is useful for? Also, what is the margin of error with the real count? Thanks.

To answer your question, the `estimateTotalHits` is the same as the old `nbHits`, we opted to rename the key because the users were induced to think this was the total amount of documents (you can read more about [here](https://github.com/meilisearch/product/discussions/483)).

We're planning to provide a new way of pagination in the next Meilisearch versions, but this is still ongoing work https://github.com/meilisearch/meilisearch/issues/2577.

jmarsh24:
Is there a workaround for this issue until it's patched?

benbws:
As per brunoocasali's answer on the related issue #394 "I'll release a new version of the ruby gem tomorrow that will fix this issue. In fact, you can try right now if you want to https://github.com/meilisearch/meilisearch-ruby/pull/349" - I'm going to close this pull request as it is not required.

brunoocasali:
The new version of meilisearch-ruby is out: https://github.com/meilisearch/meilisearch-ruby/releases/tag/v0.19.1

PR #403, opened 2022-08-14
Allowing users to override the request path

elliotlarson:
references: https://github.com/ddnexus/pagy/discussions/402

### DIFF
--- a/lib/pagy.rb
+++ b/lib/pagy.rb
@@ -22,9 +22,10 @@ class Pagy
               fragment:   '',
               link_extra: '',
               i18n_key:   'pagy.item_name',
-              cycle:      false }
+              cycle:      false,
+              request_path: '' }
 
-  attr_reader :count, :page, :items, :vars, :pages, :last, :offset, :in, :from, :to, :prev, :next, :params
+  attr_reader :count, :page, :items, :vars, :pages, :last, :offset, :in, :from, :to, :prev, :next, :params, :request_path
 
   # Merge and validate the options, do some simple arithmetic and set the instance variables
   def initialize(vars)
@@ -34,6 +35,7 @@ class Pagy
     setup_pages_var
     setup_offset_var
     setup_params_var
+    setup_request_path_var
     raise OverflowError.new(self, :page, "in 1..#{@last}", @page) if @page > @last
 
     @from   = [@offset - @outset + 1, @count].min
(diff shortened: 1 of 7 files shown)


lib/pagy.rb
```diff
@@ -116,6 +118,16 @@ def setup_params_var
     raise VariableError.new(self, :params, 'must be a Hash or a Proc', @params) \
           unless (@params = @vars[:params]).is_a?(Hash) || @params.is_a?(Proc)
   end
+
+  def setup_request_path_var
+    request_path = @vars[:request_path]
+    return if request_path.blank?
+
+    raise VariableError.new(self, :request_path, 'bust be a bare path like "/foo"', request_path) \
+          if !request_path.start_with?('/') || request_path.include?('?')
```
elliotlarson, line 127: I figure it would be pretty easy to use something like `foos_url` instead of `foos_path` when trying to set the request path with Rails URL helpers.  It seems like it would be useful to protect against fully qualified URLs with domain names and URLs or paths with additional params.

ddnexus: Small typo ("bust" / "must")

lib/pagy/url_helpers.rb
```diff
@@ -8,6 +8,7 @@ module UrlHelpers
     # For non-rack environments you can use the standalone extra
     def pagy_url_for(pagy, page, absolute: false, html_escaped: false)
       vars                = pagy.vars
+      request_path        = vars[:request_path].presence || request.path
```
ddnexus, line 11: `presence` is rails, not ruby!

lib/pagy.rb
ddnexus: `blank?` is rails not ruby!

ddnexus (commented):
Very fast and very good overall. Thank you!

I noticed that you didn't add any entry in the `pagy.rb`, which is good, since it does not make a lot of sense as a global default.

We should release this for version 6, (v6 branch) because it's not so urgent and the doc will require to be moved to the new docs structure. (Also the url helper is slightly different in v6).

@elliotlarson if you still have time, it would be great if you can move it to the branch v6. If not, @benkoshy or I can do it later.

benkoshy:
BTW since you mentioned using streams - there's a [PR in Turbo](https://github.com/hotwired/turbo/pull/612) (relatively recently) merged on the beta release which allows for streams on GET requests:

```ruby
<%== pagy_nav(@pagy, link_extra: 'data-turbo-stream="true"') %>
```

And then all you need do is have the end point stream the results back:

```ruby
<% turbo_stream.replace "foos" do %>
    <% @foos.each do |foo| %>
        <%= render  foo %>
    <% end  %>

    <%== pagy_nav(@pagy, link_extra: 'data-turbo-stream="true"') %>
<% end  %>
```
(Untested). But am curious to see your implementation.

I can merge it no problem - all it needs is a very few minor touch ups and you have won the day. I will ensure the documentation is on the retype branch. You are the champion:

we_are_the_champions


(This incidentally makes auto incremental navigation much easier........especially if combined with some turbo frames chicanery)

lib/pagy.rb
elliotlarson: Ah, right!  I saw ActiveSupport in the Gemfile but didn't realize that was only in the test environment.  I'll update this.

elliotlarson:
@ddnexus Okay, I've added in some updates based on your review.  I'm going to close this PR in favor of #404 based on the V6 branch.

Issue #410, opened 2022-11-14
`ArgumentError` when using `pagy_bootstrap_nav_js` with `countless` under Ruby 3


hvanhonacker:
# Code Issue
- [x] I researched through the [documentation](https://ddnexus.github.io/pagy/), the [pagy issues](https://github.com/ddnexus/pagy/issues) and a known Search Engine, and there is no reference|report|post|video that solves this problem
- [x] I did install|upgraded to the latest version of pagy (or the latest `3.*` for pagy legacy)
- [x] I am providing at least one of the following working and self-contained code support that can reproduce this issue:
   
   - [ ] Simple step by step list that would work in IRB with the [Pagy::Console](https://ddnexus.github.io/pagy/api/console)
   - [ ] Plain ruby file that can run as `ruby my-problem.rb`
   - [ ] Edited copy of the single file [pagy_standalone_app.ru](https://github.com/ddnexus/pagy/blob/master/apps/pagy_standalone_app.ru)
   - [x] Edited copy of the single file [pagy_bare_rails.rb](https://github.com/ddnexus/pagy/blob/master/apps/pagy_bare_rails.rb)
   - [ ] Link of my own branch forked from pagy, which contains an added test file  
   - [ ] Link of my own branch forked of [any of the rails apps listed here.](https://github.com/stars/benkoshy/lists/rails-demo-apps-for-pagy)
   - [ ] Link of a standalone `docker image` downloadable from a docker repository (e.g. dockerhub.com)
   - [ ] `docker-compose` file that can be run with `docker-compose up` (and doesn't need to be built because it uses only docker images and no local context)

### Versions

- Ruby 3.1.2p20
- Rails 6.1.7
- Pagy 5.10.1

### Description

Under Ruby 3, when using `pagy/extras/bootstrap` and `pagy/extras/countless`, calling `pagy_bootstrap_nav_js` raises the following error:

Using Ruby 3: 

```
ArgumentError (wrong number of arguments (given 1, expected 0)):
  
lib/pagy.rb:47:in `series'
```

### Steps to reproduce the issue using the provided support

Using Ruby 3, run: 

```ruby
# frozen_string_literal: true

# Self-contained, bare rails app, usable to play with pagy
# and/or easily reproduce any pagy issue.

# Edit this file as you need

# USAGE:
#     ruby pagy_bare_rails.rb

# NOTICE: if you get any installation error(s) with the following setup
[...]

PR #411, opened 2022-11-14
Fix the args forwarding in the Pagy::Countless#series override for Ruby 3

hvanhonacker:
## Description

Fixes https://github.com/ddnexus/pagy/issues/410

👋  when `extras/countless` is used in combination with `extras/bootstrap` under Ruby 3 and the method `#pagy_bootstrap_combo_nav_js` is called, an error was raised in lib/pagy.rb:47:in `series':

`ArgumentError: wrong number of arguments (given 1, expected 0)`

This is due to how Ruby 3 separates positional and keyword args. This fix allows to forward properly both kinds of args to `super` in `Pagy::Countless#series`


ddnexus:
Edited and merged in 02013d426fdd56c366332fc450bd63197a86762c

PR #412, opened 2022-11-25
Remove aria-label="pager"

sunny:
The `aria-label` attribute should only be added to interactive elements (`a`, `button`, etc) and therefore not on the `<nav>`.

See also https://ericwbailey.website/published/aria-label-is-a-code-smell/

From reading [this commit’s title](https://github.com/ddnexus/pagy/commit/f5e3482) I think this might have already been considered but not actually changed.

Also, some minor RuboCop fixes that are not related to this change but that prevented RuboCop from being green on the build.


ddnexus:
Love it! Thank you @sunny !

sunny:
Thank you for maintaining the speedy paginator \o/

PR #417, opened 2022-12-13
Add Meilisearch finite pagination by default in the meilisearch extra

brunoocasali:
Hi @ddnexus, this is the implementation to v6, which will support the finite pagination of Meilisearch out of the box.

The `limit` and `offset` params should no longer be used in this context, so this will be the most significant breaking change.

- [x] Try out with a rails app.

### DIFF
--- a/docs/extras/meilisearch.md
+++ b/docs/extras/meilisearch.md
@@ -20,7 +20,7 @@ require 'pagy/extras/meilisearch'
 If you have an already paginated `Meilisearch` results, you can get the `Pagy` object out of it:
 
 ```ruby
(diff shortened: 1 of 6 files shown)


ddnexus:
@benkoshy Please, when the retype branch will be updated, could you update the retype doc with the doc changes in this PR. Thanks!

ddnexus:
@brunoocasali There is a meilisearch related question, asked months ago (#400) that you may want to address officially...

brunoocasali:
I tried locally with a simple rails app, and it seems to be working fine @ddnexus :)

ddnexus:
Thank you @brunoocasali !

PR #428, opened 2022-12-22
Add latest version number for v5 to the CHANGELOG to help transition to v6

sunny:
Since I had to look it up, it might as well be written explicitly in the path to v6.


ddnexus:
True, thank you!

PR #429, opened 2022-12-22
docs: add warning note re: overriden method

benkoshy:
Add a note suggesting people check whether they have properly over-ridden the `pagy_url_for` method


ddnexus:
Thank you!

PR #434, opened 2022-12-27
docs: simplify introduction

benkoshy:
### What changed?

- intro is significantly shorter (41% shorter)
- the wording is more precise, certain parts omitted given it's a pure introductory passage.
- dot points used when enumerating.

### Why?

* shorter is sweeter.

I hope it helps. if not, feel free to discard.

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,42 +6,41 @@ icon: home
 
 ## Pagination is a simple task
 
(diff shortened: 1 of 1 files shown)


ddnexus:
hey Ben, thank you for this. 
Your version is certainly better than what it was before, however the content is really not what we need, and this PR pointed that out.
The repo README file is designed to be a home page, so be it. I included it in the docs as it is in the repo, and it works a lot better.

Thank you!

PR #453, opened 2023-01-26
docs: update bootstrap images

benkoshy:
Why?

 - old images are not applicable as per: discussion: https://github.com/ddnexus/pagy/discussions/444 because the middle button was disabled, but now the color is grey-ish.

 - Also, old images had two combinations "white" and "gray" that was due to the old docs site, which is no longer necessary given we're on retype. These images have been deleted and updated.

 - add short caption to explain what is going on.

 - add image for standard bootstrap_nav to prevent confusion with the other images existing below.


docs/extras/bootstrap.md
```diff
@@ -59,13 +61,13 @@ The `bootstrap_nav.*` templates produce the same output, and can be used as an e
 
 === `pagy_bootstrap_nav_js(pagy, ...)`
 
-![bootstrap_combo_nav_js](/docs/assets/images/bootstrap_nav_js-g.png)
+![bootstrap_nav_js - Notice how it gets shorter, depending on width?](/docs/assets/images/bootstrap_nav_js-g.png)
```
ddnexus, line 64: It's just me, but could we just write something without questions like "bootstrap_nav_js - It can responsively adjust its :size to the available width"?

docs/api/javascript.md
```diff
@@ -41,7 +41,7 @@ See [Pagy Navs](javascript/navs.md)
 
 +++ `pagy*_combo_nav_js`
 
-![bootstrap_combo_nav_js](/docs/assets/images/bootstrap_combo_nav_js-g.png)
+![bootstrap_combo_nav_js](/docs/assets/images/bootstrap_combo_nav_js.png)
 
 * Navigation and pagination info combined.
```
ddnexus, line 47: This file uses also the nav_js image, so we should change the image to the white version, adding also the same alt description.

docs/extras/bootstrap.md
benkoshy: > we should use the white background version.

Shall i get rid of the grey versions? I don't think we need them anymore (?)

ddnexus: > Shall i get rid of the grey versions?

Yes please. I thought you were doing exactly that already. I completely forgot about the background when I worked on the new doc.

benkoshy: My pics are white anyway (or should be close enough to white). I'll change the name of the files to reflect this to avoid confusion. apologies for it.

ddnexus: we will have only one background so no need to rename and adding the abbreviation in the name.

benkoshy: > It's just me, but could we just write something without questions like "bootstrap_nav_js - It can responsively adjust its :size to the available width"?
> 
> Also: we should use the white background version.

fixed with: [92ece79b](https://github.com/ddnexus/pagy/pull/453/commits/92ece79b6769d19a80bf77b4075425f470c46391)

benkoshy: Re: Renaming of images:

fixed with [7aa46b96](https://github.com/ddnexus/pagy/pull/453/commits/7aa46b9664e9aa2db378b219aeaf9990356c1167)

docs/api/javascript.md
benkoshy: fixed with [b8bf632](https://github.com/ddnexus/pagy/pull/453/commits/b8bf6328d00d2d23cf028455247035edddf7139c)

benkoshy:
(If you're happy, i'll rebase and force push.)

docs/extras/bootstrap.md
```diff
@@ -61,7 +61,7 @@ The `bootstrap_nav.*` templates produce the same output, and can be used as an e
 
 === `pagy_bootstrap_nav_js(pagy, ...)`
 
-![bootstrap_nav_js - Notice how it gets shorter, depending on width?](/docs/assets/images/bootstrap_nav_js-g.png)
+![bootstrap_nav_js: It is responsive.](/docs/assets/images/bootstrap_nav_js-g.png)
```
ddnexus: Then I would prefer "Responsive nav".

ddnexus: ... and no need to force-push, because Github can do it by itself on merge.

benkoshy: fixed with: [3605d15](https://github.com/ddnexus/pagy/pull/453/commits/3605d157420478e68f41fa50711f1950adadfc5a)

benkoshy:
> This looks OK to me. Should I wait for any other change?

No more changes; you may squash and merge etc.

ddnexus:
Thank you!

PR #454, opened 2023-01-26
docs: add images for semantic

benkoshy:


ddnexus:
It looks like the images have random pixels in the background coming from the page of the screenshot.

If possible I would also cut the screenshots with the same margins around the edges (it is quite visible with dark doc), also trying to make them consistent between images.

How the nav_js screenshot shows that it's responsive? We should add an alt description like the one for bootstrap, besides why should we use a different image than the one for the plain nav's?

benkoshy:
* Fix: background pixels - remove them. (nicely spotted, I didn't notice them until you poitned it out).
* Fix: standardized margins: all done to 10px (via photoshop)

> How the nav_js screenshot shows that it's responsive? We should add an alt description like the one for bootstrap.

Good point. Perhaps it might be worth adding a demo of the responsiveness: i can do so in a separate PR, but I would need to fix pagy_rails first.

ddnexus:
> Good point. Perhaps it might be worth adding a demo of the responsiveness: i can do so in a separate PR, but I would need to fix pagy_rails first.

So better to just remove the image for now.

ddnexus:
Thank you!

PR #456, opened 2023-02-06
Docs: add bulma images

benkoshy:
### Why?

- user experience: so people can see how the nav helpers look.

### What's changed:
- add images / rename.
- add 10 px margin, so it fits nicely with the docs, esp in dark mode.


ddnexus:
Could we have a more compact screenshot of the nav? It is very wide with a lot of empty space.

benkoshy:
> Could we have a more compact screenshot of the nav? It is very wide with a lot of empty space.

Sure I can do a compact one. I get rid of the desktop version or do you want both?

ddnexus:
The desktop version is the one to use, on a single line, just with less empty space between prev/next buttons and pages.

benkoshy:
> The desktop version is the one to use, on a single line, just with less empty space between prev/next buttons and pages.

The following:

```erb
<div class="container">
   <%== pagy_bulma_nav(@pagy) %>
</div>
```
...gives this result:

image

I can edit the `bulma_nav` helper itself to reduce the space quantity as a default..... besides that: I would not know how to reduce the space without adding further markup? 

What do you think?

> Also, on a more general note valid for all helpers, I think we should show the panels always open when we have a sample image for the helper.

Agreed, I can do this.

ddnexus:
> I would not know how to reduce the space without adding further markup?

Shouldn't just manually resizing the browser window before taking the screenshot do that?

benkoshy:
> > I would not know how to reduce the space without adding further markup?
> 
> Shouldn't just manually resizing the browser window before taking the screenshot do that?

That's correct, but with a caveat: the screenshot differs depending on window size - i will never be able to get it with zero spacing without additional markup:

**Minimum spacing possible**:

the minimum spacing that can be done.

If I reduce the window smaller any smaller, than this is the result:

**Medium:** 

 Medium


The nav differs a lot depending on window size:

**Full:**

Full

**Compact:**

compact


What do you think?

ddnexus:
Medium will be fine. Thank you

ddnexus:
Thank you!

PR #458, opened 2023-02-08
docs: open applicable retype panels by default

benkoshy:
What is this?

* Panels with images within them are closed. They should be opened, so users can see what is inside. This PR addresses this.

image


ddnexus:
Thank you!

Issue #460, opened 2023-02-08
Docs: Arel extra is missing from pagy.rb


Birdie0:
### Before submitting...

- [X] I searched through the [Documentation](https://ddnexus.github.io/pagy/)
- [X] I searched through the [Issues](https://github.com/ddnexus/pagy/issues)
- [X] I searched through the [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)

### Description

Configuration file [pagy.rb](https://github.com/ddnexus/pagy/blob/master/lib/config/pagy.rb) is missing mentioning of [Arel Extra](https://ddnexus.github.io/pagy/docs/extras/arel/).

benkoshy:
Thank you for your contribution!

Closing in light of #462. Please re-open if I have missed the mark.

Birdie0:
Thanks! Maybe also add `See https://ddnexus.github.io/pagy/docs/extras/arel` line for consistency?

PR #462, opened 2023-02-08
docs: add comment - arel extra in pagy.rb config

benkoshy:
Why?

* Solves: https://github.com/ddnexus/pagy/issues/460 
* A better user experience.

### DIFF
--- a/lib/config/pagy.rb
+++ b/lib/config/pagy.rb
@@ -38,6 +38,7 @@
 # Backend Extras
 
 # Arel extra: For better performance utilizing grouped ActiveRecord collections:
+# See: https://ddnexus.github.io/pagy/docs/extras/arel
 # require 'pagy/extras/arel'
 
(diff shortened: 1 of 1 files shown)


lib/config/pagy.rb
```diff
@@ -37,6 +37,9 @@
 
 # Backend Extras
 
+# Arel extra: For better performance utilizing grouped ActiveRecord collections:
+# require 'pagy/extras/arel'
```
ddnexus, line 41: While at it, please, add the reference to the do
```suggestion
# See https://ddnexus.github.io/pagy/docs/extras/array
```

benkoshy: Good point. I missed that.

benkoshy: Fixed with: [349b15e](https://github.com/ddnexus/pagy/pull/462/commits/349b15eaf868386db660f81e38fbba4458cde7e5)

Issue #470, opened 2023-02-22
Docs: Add better description for nav_js blocks


ddnexus:
Add something like:

"It looks exactly like the `pagy_<framework-name>_nav`, but it's rendered on the client side with optional responsiveness."

before the link to "See Javascript Navs" in all the `pagy_<framework>_nav_js` blocks.

benkoshy:
Closing in favour of #472.

Issue #422, opened 2022-12-21
Docs: Add responsive nav_js images


benkoshy:
### Responsive Navs

Add in responsive nav_js images for the following extras:
- [x] bootstrap
- [ ] bulma
- [ ] foundation
- [ ] materialize 
- [ ] semantic 
- [ ] tailwind 
- [ ] uikit

ddnexus:
I thought we might want to pass on this one, especially because after #470 it would probably be enough. However, having a specific image capturing the responsiveness for each CSS would not hurt.

benkoshy:
Seeing it was a low priority, we cut the scope to use the bootstrap responsive nav image everywhere - even to demonstrate how pagy looks in non-bootstrap css frameworks. This is less than ideal, but serves the purpose of demonstrating how a responsive nav looks. Instead of taking  (~4 hours photo-shopping) an acceptable outcome was achieved in less than 15 minutes. Closing in favour of #473.

PR #473, opened 2023-02-23
docs: add responsive nav image

benkoshy:
### What is this?

* Add generic responsive nav image.

### Why?

Allows for the point to get across: here's roughly what a responsive_nav_js does, without spending time customizing the repsonsive nav images for every single framework. Less than ideal, but it's quick and easy, in response to #422, and #470. Perhaps it will suffice for our purposes?

Your thoughts?


ddnexus:
If a generic image is enough, then we could just show the bootstrap one (which at least shows the real thing), with a caption that says something like "_Bootstrap style shown (the actual \<framework-name\> style looks exactly like the `pagy_<framework-name>_nav`_".

benkoshy:
Agreed that might work better.  I was concerned that people would expect the bootstrap style,  but a note might be sufficient.  I'll post the fox tomorrow.

ddnexus:
Thank you

PR #484, opened 2023-03-13
docs: fix items selector image

benkoshy:
### What is this?

* Current image had a title in it, which was superfluous.
* We've removed the title, and added a 10px margin. Same standard for all other images.


ddnexus:
Thank you!

PR #486, opened 2023-03-15
docs: add git graph

benkoshy:
### What is this?

A picture says a thousand words. I am hoping this clarifies rather than confuses.

See a demo here: https://github.com/benkoshy/pagy/blob/docs-add-git-graph/.github/chore-merge.md

### DIFF
--- a/.github/chore-merge.md
+++ b/.github/chore-merge.md
@@ -1,11 +1,87 @@
 # Chore PR merge
 
-The `admin` branch is the branch used for chores. It fast-forwards to `master` when `master` gets new commits. It receives chore PRs, and `master` fast-forwards to it when the a chore PR is merged.
+Consider the following git graph:
+
+```mermaid
+gitGraph
+    branch master
+    checkout master
+    commit id: "ZERO"
+    commit
+    branch admin
+    checkout admin
+    branch chore-pr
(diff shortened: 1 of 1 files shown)


ddnexus:
Merging to admin not used anymore.

benkoshy:
So we'd merge directly. Steps:

[ ] So just check all the tests pass.
[ ] No need to squash and rebase? Just normal merge.

Please confirm?

ddnexus:
None of that 😉

We could just merge (Merge button in the PR) and we would have an extra merge commit per PR, regardless the status of the PR compared to the current master HEAD. That could become cumbersome when there are multiple chore PRs and eventually more commits in master created after the chore PRs.

However, if we want a simpler and linear history, would should always use `@dependabot rebase` inside of each chore PR (it will do nothing in case it would not be needed), just before using the Squash and merge button. That would ensure that the PR commit will be added as a single commit on top of the current master.

If the tests fail, then manually revert the offending merge (which will be the last one).

ddnexus:
So the optimal steps are:

1. Check the tests pass
2. `@dependabot rebase`
3. Check rebase completes (one way or another)
4. If rebased, check tests pass
5. `Squash and merge`

Issue #405, opened 2022-08-18
Is there a configuration to only show pagination when there is more than one page?


evgenyneu:
Hi, is there a Page configuration to only show pagination with `pagy_nav(@pagy)` when there is more than one page? Of course, I can manually do this:

```
<% if @pagy.pages > 1 %>
    <%== pagy_nav(@pagy) %>
<% end %>
```

but I was wondering if this is already built into `pagy_nav`.

ddnexus:
Manually by design. Details [here](https://ddnexus.github.io/pagy/how-to#skip-single-page-navs)

benkoshy:
It's the only way:

You can try something like this if you wish.

```rb
<% if @pagy.pages > 1 %>
    <%== pagy_nav(@pagy) %>
<% else %>
    <%== pagy_info(@pagy) %>
<% end %>
```

(Typically we reserve issues for bug fixes. It doesn't sound like a bug so I would probably close this issue).

PR #491, opened 2023-03-20
docs: add canonical example

benkoshy:
Close: #405, also I've seen this question before.


ddnexus:
It is addressed [here](https://ddnexus.github.io/pagy/docs/how-to/#skip-single-page-navs) and adding it to he readme (and in that multi-line form) is adding too much distracting bulk to the readme.

PR #490, opened 2023-03-20
docs: add how-to: paginate with indepedent contexts

benkoshy:
Supported with an example case here: https://github.com/benkoshy/pagy-rails/commit/dff6994a206cc2379a31d212d2876fbaa92950d9

PR #495, opened 2023-03-21
Docs: add independent context example

benkoshy:
Same as in #490.


ddnexus:
What about including the "Customize the request path" section to this context?

I would sub-section it with "Customize the request path", "Hotwire/Turbo Frame", "Multiple models" and move it all after the "Paginate by date..." section.

Issue #471, opened 2023-02-22
Docs: Update AJAX doc


ddnexus:
Consider updating [AJAX](https://ddnexus.github.io/pagy/docs/api/javascript/ajax/) for the modern day tooling - nobody uses .js.erb anymore - Turbo Streams (and Stimulus JS) are what they use, and it serves much the same purpose.

benkoshy:
### Supporting Code?

Most people who AJAX Render would need to re-initialize Javascript - somehow: an efficient way of solving this problem is via [Stimulus Controllers](https://stimulus.hotwired.dev/reference/controllers) - which allows for javascript to be initialized as soon as it hits the DOM. We have already alluded to this here: https://ddnexus.github.io/pagy/docs/api/javascript/setup/#stimulus-js

* Does it make sense to add this code to the pagy library?
* Or should we merely document it?
* Or should we produce an installer to dump it in the appropriate directory?
* I imagine it would need very little support, given there are almost zero moving parts. But it will still be there.

### AJAX rendering

I feel like i'm not quite understanding the purpose of this AJAX section.

* user has a `js_nav` helper.
* user clicks on page 2.
* new records are fetched and the `nav_js` is re-rendered in the same controller action (or are pages rendered in one action, and the pagination component rendered in a different action)?
* the `nav_js` helper is re-initialized.

ddnexus:
> Does it make sense to add this code to the pagy library?

Nope. Too many different ways that may just be popular for a few months, depending on the last trend. It's already difficult to keep up with the doc.

> Pls confirm whether this is incorrect (or correct)?

The click on page 2 requests the same action, unless you use `:request_path`.

ddnexus:
As a general policy, we want to make customization always possible. Hopefully also easy.

When it's already easy - as in the javascript case - we can avoid adding (bloating) more code base, that will need more testing and maintenance without any real need.

benkoshy:
> The click on page 2 requests the same action, unless you use :request_path. BTW, the url is specifically generated by the pagy_url_for method.

Turbo drive works via AJAX, which most Rails folks use by default. Turbo stream / Frames targets specific dom elements, and renders `.js.erb` obsolete, so all that really needs to be done is to re-initialize the js helper. 

This issue and  #451 could probably be addressed the same. i will ponder on how to structure it.

anitagraham:
I'd agree @ddnexus  -  many/most is not all.
[...]

PR #497, opened 2023-03-23
Docs: add canonical turbo-stream example using ransack

benkoshy:
Update: not ready to merge.

### Why?

* To address this recurring issue: https://github.com/ddnexus/pagy/discussions/450
* To address the new way AJAX is being managed. #471

What I have not addressed / discussed:

* streaming using Post requests. This is currently covered by a blog post.

I have directly extracted from [pagy-rails](https://github.com/benkoshy/pagy-rails), which itself is an extraction from my experiments with turbo. 

There are many ways you can return results: using turbo_frames or streams. I'm not sure which is the "correct" way -  both can be used.

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -418,6 +418,8 @@ Ransack `result` returns an `ActiveRecord` collection, which can be paginated ou
 ```
 |||
 
+Review the following link if you wish to [turbo stream the response with a GET request](#paginating-while-using-turbo-streams).
+
 ## Paginate search framework results
 
 Pagy has a few of extras for gems returning search results:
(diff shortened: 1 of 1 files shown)


docs/how-to.md
```diff
@@ -519,6 +521,83 @@ If you're using [hotwire](https://hotwired.dev/) ([turbo-rails](https://github.c
...
+import throttle from 'lodash.throttle'; // yarn add lodash.throttle
+
+// lodash and debounce help limit the rate 
+// at which requests hit your server.
+
+export default class extends Controller {
+  static targets = ["submitButton"]
+
+  connect(){        
+    this.submit = debounce(this.submit, 250).bind(this)
```
yshmarov, line 585: just wondering, what is the benefit of using a debounce library rather than just this:
```js
  debounce() {
    clearTimeout(this.timeout)
    this.timeout = setTimeout(() => {
        this.formTarget.requestSubmit()
      }, 500)
  }
```

benkoshy: absolutely you can do that too.

PR #509, opened 2023-04-14
Docs: add tooltips

benkoshy:
### Why?

* So people can get a fuller explanation of what something does / is, without navigating off screen.

### Costs?

* Negligible: nothing in the presentation is effectively changed: the tooltips only becomes visible on hover. It effectively gives links a 'title'. This will affect those with a visual impairment in a (hopefully) positive way: there is no need to further navigate to a different page - they can get a brief description on the readme itself.


ddnexus:
This is simple and very useful!
Let's add tooltips to as many links as possible. 

For example, we could add some comment/description about the kind of Screencast/post we link. Even just the duration and the skills required would be helpful, but there might be other simple info that could save time to the users.

benkoshy:
> For example, we could add some comment/description about the kind of Screencast/post we link. Even just the duration and the skills required would be helpful, but there might be other simple info that could save time to the users.

I'm on board with this.

Any issues if we merge several commits to master as they are done? My preference is to avoid long lived branches off master - i.e. to keep master as close to the updates as possible: mainly because i want to avoid the risk of merge conflicts: delivering snippets of usefulness to users in over time vs one big batch?

ddnexus:
> avoid long lived branches off master

In principle, yes, but in this particular case, we only change a single readme file without any code, which makes conflicts unlikely or very easy to resolve.

PR #511, opened 2023-04-17
Docs: remove pics in lieu of shorter statements

benkoshy:
Why?

* to shorten read me.
* also, pics are already listed below - they weren't before. I feel this section can safely be elided.
* links with titles provided in lieu.

Just suggestions. Such views always involve weighing pros / cons.


ddnexus:
I like the removal of the pix that are already shown elsewhere. 
However, please keep the same format as the paragraph above:

- "Pagy" in the first topic, not in the headline
- Bold "titleized" (first upper-cased char)

benkoshy:
I have reverted the original text.

reverted_original_text

PR #516, opened 2023-05-02
fix: Extras::Trim - fix trimming first page

tulak:
Multiple frontend extras construct the navigation links by looping through the elements of `pagy.series` and passing the elements to the `lambda` (previously returned by `pagy_link_proc`). The API of `pagy.series` seems to follow the convention that:

- if it's an `Integer` it's a link to the page
- if it's a `String` it's a link to the current page
- if it's a `:gap` it's a gap element.

Some of the frontend extras (`FoundationExtra`, `NavsExtra`) do not render the link when rendering a current page. But some (`BootstrapExtra`,`MaterializeExtra`, ...)  [call](https://github.com/ddnexus/pagy/blob/cf0dae2bcb4564f75f42b70c5a948ffebce5df42/lib/pagy/extras/bootstrap.rb#L22) the `lambda` even with `String` element as a page identifier.

This PR fixes the `TrimExtra` to also perform trimming if the identifier of the page is `'1'` (first page which is also current page)


ddnexus:
Good catch! Thank you!

tulak:
Thank you @ddnexus  for updating the e2e snapshots. I was struggling with setting up Cypress a bit.

ddnexus:
You are welcome!

BTW Didn't the [README](https://github.com/ddnexus/pagy/tree/master/e2e/README.md) help?

PR #518, opened 2023-05-08
Docs: simplify warning re: pagy_array

benkoshy:
Why?

Fix for: https://github.com/ddnexus/pagy/discussions/514

Looks like too many people are misusing this extra.  I think the current warning was not obvious enough? I thought why not shorten the warning and make it really obvious with red danger signs: 

do not do

### DIFF
--- a/docs/extras/array.md
+++ b/docs/extras/array.md
@@ -9,43 +9,24 @@ categories:
 
 Paginate arrays efficiently.
 
-!!! warning WARNING
-The `array` extra is efficient if you need to paginate an array, but if the data in the array comes from some DB or other persisted storage (i.e. not some in-memory storage), DO NOT use the array extra.
-!!!
-
-+++ Bad
-!!!danger Do not:
+!!!danger Bad - do not do this:
 ```rb
 def index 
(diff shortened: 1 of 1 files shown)


ddnexus:
> the current warning was not obvious enough?

It is more than obvious IF you read it. Maybe it's  challenging because it's more than 280 characters long? 🤭

ddnexus:
So shortening it further, approaching the golden 280 character length ... instead of good/bad examples, or the old warning, I would just write:

Don't use (warning)
With any persistent storage collection (e.g. database, elastic search, meilisearch, etc.)

Use (success)
With collections that are already loaded in memory for whatever reason. (e.g. arrays of cached indices, keys of hashes, pointers, etc.).

No tabs with that simple messages.

benkoshy:
I agree. I think you are [100% bang on the money](https://en.wiktionary.org/wiki/bang_on_the_money). 

If it is longer than a tweet, I don't think people actually read it. 

I will make the changes that have been suggested.

benkoshy:
I adopted the above suggestions with some minor changes. 

I hope this warning will have it's desired effect.

PR #533, opened 2023-06-16
Update overflow valid mode :empty_page

xuanxu:
The listed `:empty` option is not a valid mode. This PR updates the docs to show the correct `:empty_page` value.


ddnexus:
Thank you!

PR #540, opened 2023-07-30
Fix: links to docs

espen:
These links are 404. They should either be .md or as in this commit with link to the pagy website.


benkoshy:
Nice one! Thank you for posting the fix.

I would wish to retain the relative links, rather than hardcoding the path.

I will make the changes based upon your commit.

ddnexus:
> I would wish to retain the relative links, rather than hardcoding the path

The readme must work from 2 domains, so it can't be relative. 🤷

benkoshy:
Close in favour of  #542 - your commit has been amended.

Thank you for your PR!

PR #542, opened 2023-07-31
Fix: links to docs

benkoshy:
These links are 404. They should either be .md or as in this commit with link to the pagy website.
-- @espen 

Agreed: relative links used with .md extension. Why?

* I would want the docs to work from many domains e.g. from your own fork of github pages, or from a development environment. hard-coding to the main repo, while it does work, I feel is somewhat restrictive.
* You may test out the links here: https://benkoshy.github.io/pagy/

I have amended your commits, so they will be here for posterity to see. Or you may alternatively re-submit with the same commits. 

If I've missed the mark, feel free to re-open. Thank you very much for the PR!

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,14 +113,14 @@ Or, choose from the following view helpers:
 
 | View Helper Name                                                                                                                                                                        | Preview (Bootstrap Style shown)                                        |
 |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------|
-| [`pagy_nav(@pagy)`](/docs/api/frontend)                                                                                                                                                 | ![`pagy_nav`](/docs/assets/images/bootstrap_nav.png)                   |
-| [`pagy_nav_js(@pagy)`](/pagy/docs/api/javascript/)                                                                                                                                      | ![`pagy_nav_js`](/docs/assets/images/bootstrap_nav_js.png)             |
-| [`pagy_info(@pagy)`](/docs/api/frontend)                                                                                                                                                | ![`pagy_info`](/docs/assets/images/pagy_info.png)                      |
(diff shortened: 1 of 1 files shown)


ddnexus:
The README links to `.md` files point to the right retype doc ONLY when the domain is `github.io`.

When it is `github.com`, they point to the raw file in the repo, missing context, formatting and tools available in the retype doc. 

That's why I said that we must use absolute links, unless you find a relative link alternative to get always the retype doc.

benkoshy:
> When it is github.com, they point to the raw file in the repo, missing context, formatting and tools available in the retype doc.
> That's why I said that we must use absolute links, unless you find a relative link alternative to get always the retype doc.

Ok that makes sense. I had misunderstood.

This raises an interesting question: 

- what advantage is gained from viewing the docs (pre-build) from github.com? I imagine most people would view the retype docs (after they have been built) as opposed to reading them on github.com where the formatting etc. is lost? I would suggest we develop the docs with `retype watch` or `retype start` and let the links be broken when viewing from github (which is not very readable anyway). Perhaps I have misunderstood? PLMK your thoughts?

ddnexus:
> I would suggest we develop the docs with retype watch or retype start and let the links be broken when viewing from github (which is not very readable anyway).

The whole point is that the github.com README is the landing page for most users so "let the links be broken" doesn't strike me like a good idea.

benkoshy:
> The whole point is that the github.com README is the landing page for most users when they access the repo, so "let the links be broken" doesn't strike me like a good idea.
> 
> We want the links to work AND we want them to switch the browser to the github.io retype context when used from github.com (AND of course we want to use the exact same README file in both domains).
> 
> That's only for the README file. All the other files are developed in the retype context (i.e. using retype special formatting etc.). Also the README file must not contain any special formatting, because githum.com would not understand it.

Ok very good. 

I had misunderstood the context and your explanation makes it perfectly clear. 

In that case we may close this PR and re-open the one proposed by @espen?

PR #554, opened 2023-09-05
Fix: broken links

benkoshy:
The links are within code comments. There is no change to functionality.

### Why is it needed?

* When adding locales: to make it easier to find the relevant files.


benkoshy:
Not sure why CI is failing. I'll come back to this.

Issue #562, opened 2023-10-27
Docs: No documentation on how to set items per page in tests


leewaa:
### Before submitting...

- [X] I searched through the [Documentation](https://ddnexus.github.io/pagy/)
- [X] I searched through the [Issues](https://github.com/ddnexus/pagy/issues)
- [X] I searched through the [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)

### What

The documentation should include how to configure Pagy items per page configuration within tests. The only configuration available is `Pagy::DEFAULT` which is a constant, and frozen in initialization. So it is unclear how to properly configure items per page within tests.

### Why?

This should be part of the documentation to help developers move forward quickly without having to hunt the internet for solutions. Especially when using only the minitest or rspec framework without any other gems for mocking/stubing etc. Currently it is unclear how to do this correctly and the documentation does not mention anything about testing.

Ideally such a test scenario would do the following

1 change default config to 2 items per page,
2 carry out the test
3 reset back to the default

The reasons behind this should be clear: to be able to write efficient and clear tests.

There are multiple situations where a test would include pagination, be it a full happy path feature test, system test, controller test, or even unit. 

Having a section on this with information on how to correctly configure pagination in a test (rspec/minitest) would be very useful.

ddnexus:
@leewaa there's no core code that freezes `Pagy::DEFAULT` anywhere in the pagy gem. The only line that does so is in the default `pagy.rb` configuration file, which is actually part of your own code. You are supposed to edit and add that file to your own app.

That means that you could simply avoid freezing the default variables in your own tests if you really need to do so, although it's not the best practice.

A default should be frozen exactly because it's a... default. Why would you change a default AFTER you set it? You should rather test that your code shouldn't be able to change it, since, indeed, it's a default.

If you feel the need to test the `Pagy:: DEFAULT` itself, you are likely trying to test pagy itself, which is already extensively covered (100% coverage indeed) and should be outside the scope of testing your app.

Remember that if you want to change a variable, you can just set it's value in any constructor (overriding the default). That can be part of your tests, but it makes sense only if your code change the variable dynamically.

That would be evident if you use a coverage gem, that I strongly recommend. If your testing is coverage driven, you write the test to cover the (yet uncovered) code that you wrote, instead of trying to guess which test you should invent to cover the behavior of your features (which is actually determined by your code).

That way you ensure that all your code/features are covered, without wasting time in redundant mental gimmick.

[...]

PR #563, opened 2023-10-31
Docs: add testing section

benkoshy:
### Why this PR?

Fix: #562.

I added to the how-to section, because that was the best place I could think of.

(I am getting the feeling that the how-to section is getting increasingly unwieldy, but I don't have any great ideas on how to make it super streamlined for users.)


ddnexus:
Thank you!

PR #572, opened 2023-12-05
Docs: explain: pagy + retype, add contribution instructions

benkoshy:
Why? To document hidden steps - without which it might be difficult for someone to piece it altogether. also to lower the barrier to contributing to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ See also the [How To Page](https://ddnexus.github.io/pagy/docs/how-to)
 - [How To (quick recipes)](https://ddnexus.github.io/pagy/docs/how-to/)
 - [Changelog](https://ddnexus.github.io/pagy/changelog)
 - [Deprecations](https://ddnexus.github.io/pagy/changelog#deprecations)
+- [How Pagy's Docs work?](https://github.com/ddnexus/pagy/blob/master/docs/README.md)
 
 </details>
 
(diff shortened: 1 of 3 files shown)


ddnexus:
Thank you @benkoshy. A reference to doc on doc was missing, so it's a very useful addition, however I would reduce it to a single point in the list under the contributions section in the README.

If more content is really needed, please point a link to a new doc page or section, so the README will remain mostly a pointer.

Thank you!

benkoshy:
> Thank you @benkoshy. A reference to doc on doc was missing, so it's a very useful addition, however I would reduce it to a single point in the list under the contributions section in the README. (Something like "- For doc contributions ..."
> 
> If more content is really needed, please point a link to a new doc page or section, so the README will remain mostly a pointer.
> 
> Thank you!

* I have removed the section and added a link
* I have also removed the page from the retype docs. I figure it can safely be exluded without cluttering - but those who want to find it will easily see it via the link, or it will be obvious in the `docs` folder.

ddnexus:
@benkoshy I didn't lock the rubocop version in the CI Gemfile, so the CI will always use the latest, despite the one locked in the repo Gemfile.lock.

That way we are forced to notice that we should update gems and javascript modules, in order to keep the page environment always up-to-date.

ddnexus:
@dependabot rebase

benkoshy:
> @benkoshy I didn't lock the rubocop version in the CI Gemfile, so the CI will always use the latest, despite the one locked in the repo Gemfile.lock.
> 
> That way we are forced to notice that we should update gems and javascript modules, in order to keep the pagy environment always up-to-date.

 Should I update pagy to comply with rubocop? Am i understanding the implications of the above correctly?

ddnexus:
It's done. If you rebase your PR on master it should pass. :crossed_fingers:

benkoshy:
Force pushed after rebasing. 

If you're happy: squash + merge?

PR #618, opened 2024-01-15
Remove runtime dependency on base64

Earlopain:
Ruby 3.3 prints a warning if base64 is used without specifying it in the gemfile. Ruby 3.4 will error.
Since it's just a wrapper around the pack/unpack methods it's trivial to replace.

You don't see this in CI since the Gemfile contains dependencies that in turn already depend on `base64`, like `activesupport`.

Here is some similar work:
* https://github.com/rack/rack/pull/2110
* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/sidekiq/sidekiq/pull/6151
* https://github.com/opensearch-project/opensearch-ruby/pull/221
* https://github.com/lostisland/faraday/pull/1541
* https://github.com/newrelic/newrelic-ruby-agent/pull/2378

### DIFF
--- a/lib/pagy/extras/frontend_helpers.rb
+++ b/lib/pagy/extras/frontend_helpers.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'base64'
-
 class Pagy # :nodoc:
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you!

PR #625, opened 2024-01-23
Docs: add link with instructions on adding new langauges

benkoshy:
Why?

* to make this information easier to find / more transparent (I could not find it myself).


ddnexus:
Thank you @benkoshy

PR #643, opened 2024-02-15
Updates ckb translations to be complaint with ARIA in v7.x.x

muhammadnawzad:


ddnexus:
Thank you @muhammadnawzad

PR #647, opened 2024-02-19
Docs: update tailind styling for pagy + 7.0

benkoshy:
Why?

* Some html elements were not being targeted by the selector.

What changed?

* for the second styles - we need not use the href attriute selector. a simple a selector will suffice.

Couple of minor points:

* the second style targets the 'pagination' class while the first does not. It is of no consequences.
* the second style does not target nav-js styles.
* the second style: when disabled i removed the 'hover' styling.

Authored by Benjamin Block.

### DIFF
--- a/docs/extras/tailwind.md
+++ b/docs/extras/tailwind.md
@@ -46,11 +46,11 @@ simply paste the following into your `tailwind.css` file.
   @apply flex space-x-2;
 }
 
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you @benkoshy. I decided to completely refactor the style and the docs, so this one is not needed anymore.

Issue #582, opened 2024-01-10
Add cs: "pagination" entry


ddnexus:

benkoshy:
@pruzicka

Thank you again for your contribution to the cs locale.

We are in the process of improving the labelling of pagy helpers in order to comply with the [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) standards. 

We are seeking your advice with the pluralized translation of the "page", "pages" label, for each required plural listed below:

```yml
    page_label:
      one: "?"
      few: "?"
      other: "?"
```

If you are inclined: feel free to create a [PR](https://github.com/ddnexus/pagy/blob/aria/lib/locales/cs.yml#L8) or alternatively post a comment and we can manage it crediting you.

Thank you in advance for your help!

pruzicka:
> @pruzicka
> 
> Thank you again for your contribution to the cs locale.
> 
> We are in the process of improving the labelling of pagy helpers in order to comply with the [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) standards.
> 
> We are seeking your advice with the pluralized translation of the "Page", "Pages" label, for each required plural listed below:
> 
> ```yaml
>     page_label:
>       one: "?"
>       few: "?"
>       other: "?"
> ```
> 
> If you are inclined: feel free to create a [PR](https://github.com/ddnexus/pagy/blob/aria/lib/locales/cs.yml#L8) or alternatively post a comment and we can manage it crediting you.
> 
> Thank you in advance for your help!

Ha, sorry I've missed the message and I saw it seconds ago. Thanks @borama

PR #648, opened 2024-02-20
Update cs translations

borama:
Hello! This is my attempt to resolve the Czech translations needed in #582. In the end it was a bit more complex than expected so let me explain my changes a little bit more:

The problem with Czech pluralization is that the bare plural word for e.g. "items", used _without_ the numeral, is often different from the plural word when used together _with_ the numeral: It's "položky" vs. "10 položek", for example. In fact, a bare Czech plural word probably always uses the `few` variant in the i18n dictionary.

#### Aria labels

Aria label uses the pluralized word for "Pages" but in the bare form, without the numeral. That's why I used only the `other` variant in the dictionary (but with a plural form for the "few" variant of "Pages") to correctly distinguish between a single "page" and multiple "pages"  only.

#### Item_name fixes

While doing that, I noticed that the `item_name` translations were incorrect. The `few` and `other` items in the dictionary were swapped so I fixed them. 

And also, the `multiple_pages` label has to use the "bare" plural form, because it says "Items shown X-Y from Z" and not "N items shown..." and as such the `item_name` can be replaced with a static string. This still has one edge case when showing only 1 item, it will show: "Zobrazeny položky 1-1 z 1" (i.e. a plural form) instead of the more correct singular form: "Zobrazena položka 1-1 z 1" but this is not possible to do without pluralizing the word "Zobrazena" ("shown"), too. But it's not an important edge case, I guess, and it doesn't sound _that_ weird either.

### DIFF
--- a/lib/locales/cs.yml
+++ b/lib/locales/cs.yml
@@ -3,7 +3,7 @@ cs:
   pagy:
     aria_label:
       nav:
-       # skip the "few" variant as this word is used as a bare plural without the numeral
+       # use the "few" variant for all plurals as this word is used as a bare plural without the numeral
        one: "Stránka"
        other: "Stránky"
       prev: "Předchozí"
--- a/lib/locales/cs.yml
+++ b/lib/locales/cs.yml
@@ -1,8 +1,6 @@
 # :west_slavic pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/pagy/i18n.rb)
 cs:
   pagy:
-    # please add a comment in the https://github.com/ddnexus/pagy/issues/582
-    # posting the translation of the following  "Page"/"Pages" with the plurals for this locale
     aria_label:
       nav:
        # skip the "few" variant as this word is used as a bare plural without the numeral
--- a/lib/locales/cs.yml
+++ b/lib/locales/cs.yml
@@ -15,11 +15,11 @@ cs:
     gap: "&hellip;"
     item_name:
       one: "položka"
-      few: "položek"
-      other: "položky"
+      few: "položky"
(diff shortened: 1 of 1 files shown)


ddnexus:
> And also, the multiple_pages label has to use the "bare" plural form, because it says "Items shown X-Y from Z" and not "N items shown..." and as such the item_name can be replaced with a static string. This still has one edge case when showing only 1 item, it will show: "Zobrazeny položky 1-1 z 1" (i.e. a plural form) instead of the more correct singular form: "Zobrazena položka 1-1 z 1" but this is not possible to do without pluralizing the word "Zobrazena" ("shown"), too. But it's not an important edge case, I guess, and it doesn't sound that weird either.

I am confused with all that deep thinking, but I am happy that I can trust you without questioning!  😁

borama:
Haha, sorry, they say that Czech is hard and it certainly was hard to try to explain my thinking, which I still failed at, I understand :).

While updating the code, I tested all variants that came to my mind with a pager used in our project (for which I switched the locale to Czech). Let me share a few examples, just for the record:

|  | before | after |
|--------|--------|--------|
| single page message |    |  | 
| multiple pages msg |  |  |  

The top row is an example of the plural ("položky" / "položek") used together with the numeral and as such it needs all dictionary variants, `one`, `few` and `other`. The bottom row uses a bare plural and thus the plural is always the same, "položky".  I think anyone speaking Czech and seeing this will confirm that it sounds much better now :).

Thanks for this wonderful gem, btw!

simi:
As native :czech_republic: speaker, I can confirm this works well. :muscle: Thanks for the fix @borama.

PR #651, opened 2024-02-21
Fix some formatting in the readme

Earlopain:
grafik

I've moved the text for all entries below the header, the two that currently don't reder well were doing that already.The text more closely resembles the rendered output that way. To do this, appending two spaces at the end of a line forces a linebreak.


benkoshy:
Thanks for the patch: see new PR: https://github.com/ddnexus/pagy/pull/654

PR #658, opened 2024-02-28
Docs: fix broken links

benkoshy:
* The links are not actually broken.
* I could only make it trigger a download by change the style of the links. A temporary solution till a true one is found.

Try it out here: https://benkoshy.github.io/pagy/docs/api/stylesheets/

To close issue raised here: https://github.com/ddnexus/pagy/commit/ce9365daae630a6699c925437d0b96d9dc79581f

### DIFF
--- a/docs/api/stylesheets.md
+++ b/docs/api/stylesheets.md
@@ -17,7 +17,7 @@ untouched
 
 +++ pagy.scss
 
-[!file](/lib/stylesheets/pagy.scss)
+[Download file](/lib/stylesheets/pagy.scss)
 
 ```ruby 
 stylesheet_path = Pagy.root.join('stylesheets', 'pagy.scss')
@@ -27,7 +27,7 @@ stylesheet_path = Pagy.root.join('stylesheets', 'pagy.scss')
 
 +++ pagy.css
 
-[!file](/lib/stylesheets/pagy.css)
+[Download file](/lib/stylesheets/pagy.css)
 
 ```ruby 
 stylesheet_path = Pagy.root.join('stylesheets', 'pagy.css')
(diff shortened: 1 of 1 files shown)


ddnexus:
@ben, thank you for the help, but it looks far from solved.

You changed the download retype link to a regular link, which serves a page with the content of the file, instead of triggering a proper browser download.

IIRC... loooong time ago I read something in retype that says that the files should be in some asset dir in the docs dir, but we have to verify that.

ddnexus:
here is the doc https://retype.com/components/file-download/

ddnexus:
In the `quick-start.md` works like `[!file](apps/pagy_standalone.ru)`... but that is a relative path from the `quick-start.md` that is in the root dir.

so from `docs/api/stylesheets` it might work something like `[!file](../../lib/stylesheets/whatever)`

ddnexus:
Absolute links might not work, because when retype compiles itself in the docs branch, it probably looks for the linked files to copy over from its root (which is where the `retype.yml` is, which we conveniently put in the repo root).

benkoshy:
> In the `quick-start.md` it works like `[!file](apps/pagy_standalone.ru)`... but that is a relative path from the `quick-start.md` that is in the root dir.
> 
> so from `docs/api/stylesheets.md` it might work something like `[!file](../../lib/stylesheets/whatever)`

I had tried exactly this, but it failed. no idea why  ¯\\_(ツ)_/¯

I must have spent about 30-45 minutes trying to make it work (LOL) so I just gave up and posted a quick fix.

ddnexus:
It the `../../` does not work, then we might try the relative link fro retype root like `[!file](lib/stylesheets/whatever)` (if files are considered relative to its root, instead of the file that links them)

benkoshy:
> It the `../../` does not work, then we might try the relative link from the retype root like `[!file](lib/stylesheets/whatever)` (if files are considered relative to its root, instead of the file that links them)

This works in the local development environment when building retype docs. But it does not work with the file download component via github pages  ¯\\(ツ)/¯

ddnexus:
FIxed the broken link, and reconfigured retype to build the same docs-site also locally. No more surprises.
Thanks for your help.

Issue #306, opened 2021-05-02
Repeating Pagination Item


bradbajuz:
I'm having a strange issue with a repeating item when sorting by user roles.

**UserController:**

```
  def index
    @pagy, @users = pagy(User.where(role: [:supervisor, :manager, :marketer, :unassigned, :disabled]).order(:role), items: 10)
  end
```

**User Model**
```
...
 enum role: [:superadmin, :supervisor, :manager, :marketer, :unassigned, :disabled]
...
```
On my user index page, I have a HTML table with a list of users that are being ordered by the User Role.
I'm paginating like so: `<%== pagy_bootstrap_nav(@pagy) %>`

If a user with the role `unassigned` ends up at the bottom and there is more than one user that has a role of `unassigned`, the user will repeat again on the second page and one of the other `unassigned` users will be hidden.

**This ONLY happens when I order by `:role` 

### Example
I have three users that have the `unassigned` role: User1, User 14, and User 22.

**User 14 is on page one**
![user-page-1](https://user-images.githubusercontent.com/2166808/116815745-e98a5580-ab2c-11eb-8f15-e6cca6fe17a4.jpg)

**User 1 and User 14 (repeating) is on page two. User 22 is never displayed.**
![user-page-2](https://user-images.githubusercontent.com/2166808/116815747-f1e29080-ab2c-11eb-96d1-76996b2bffab.jpg)

ddnexus:
Hi @bradbajuz 

this looks like a AR count related quirk, due to the specific scope/query. Maybe related to the dreaded https://github.com/rails/rails/issues/5554.

Can you investigate the actual queries used by pagy getting the count and AR getting the scope? FYI, you can get that in the console, you will see 2 AR querires firing each time you paginate: one is getting the count, and the other is getting the actual page of records.

An easy way to do it in your console is using the [Pagy::Console](https://ddnexus.github.io/pagy/api/console) (which is a very new addition, so you may have to update pagy to the latest version).

The 2 queries are probably different, so the count returns a number, while the actual record count returned by that query is a different number.

ddnexus:
BTW, if that is the case, then you can avoid the problem by explicitly passing a `count: the_right_count` to the `pagy` method. 

See [Custom count for custom scopes](https://ddnexus.github.io/pagy/how-to#custom-count-for-custom-scopes).

bradbajuz:
I'm using Rails 6.1 and Ruby 3.0.1

Even if I simplify my query by just using:

```
[...]

PR #659, opened 2024-02-29
Docs: add warning for pgsql records ordering

benkoshy:
Refers to #306 and #546

Placed in prominent section of quick start.

image

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -1,6 +1,6 @@
 ---
 title: How To
-order: 1
+order: 8
 icon: tools-24
 ---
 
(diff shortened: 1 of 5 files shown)


ddnexus:
@benkoshy I love how it looks, however the Quick Start seems a place a bit too prominent to put it. It seems a bit like a big pharma side effect disclaimer! 😁

What about adding a Troubleshooting page somewhere (maybe just below the Quick Start), and link it at the end of the Quick Start with another nice box and a brief description... linking the Troubleshooting and the How to?

And since the warning and success boxes that you prepared look very appropriate for a problem and solution, we can keep that model in the Troubleshooting page, eventually extracting others comon problems from how-to, discussions and issues and populating that page.

Makes sense?

benkoshy:
I shortened the character count as well.

ddnexus:
Thank you!

PR #660, opened 2024-02-29
Fix formatting

lostapathy:
Removed a couple extra spaces that were breaking a link.


benkoshy:
Thanks for the patch!

PR #665, opened 2024-03-15
Fix jsonapi prev and next keys for unavailable links

kobusjoubert:
Hi @ddnexus, thank you so much for this awesome Gem 💎

I came across an issue when using the `Jsonapi` extra.

### The Problem

I'm getting a `expected :page >= 1; got 0` exception when following a `prev` or `next` link with no content.

Screenshot 2024-03-15 at 07 58 01

---

Screenshot 2024-03-15 at 08 23 50

---

### The Fix

According to the [JSON:API spec](https://jsonapi.org/format/#fetching-pagination), it specifies to omit the key or return `null` for unavailable links.

Screenshot 2024-03-15 at 08 33 12

If you're okay with this, I've applied a small change to return `null` for unavailable links in this pull request.

Screenshot 2024-03-15 at 10 51 26

---

Screenshot 2024-03-15 at 10 51 43

---

Screenshot 2024-03-15 at 10 52 01

---

Let me know if you want me to change anything @ddnexus.

### DIFF
--- a/lib/pagy/extras/jsonapi.rb
+++ b/lib/pagy/extras/jsonapi.rb
@@ -26,8 +26,8 @@ class Pagy # :nodoc:
       def pagy_jsonapi_links(pagy, **opts)
         { first: pagy_url_for(pagy, 1,         **opts),
           last:  pagy_url_for(pagy, pagy.last, **opts),
-          prev:  pagy_url_for(pagy, pagy.prev, **opts),
-          next:  pagy_url_for(pagy, pagy.next, **opts) }
+          prev:  pagy.prev ? pagy_url_for(pagy, pagy.prev, **opts) : nil,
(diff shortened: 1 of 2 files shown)


ddnexus:
Hi @kobusjoubert . Thanks for the PR.

I am conflicted about using null values or omitting... I even researched it and... it looks everybody is conflicted about null or omit. It looks like the two options are equally... meh! 🙃 

So... the option that you choose looks visually better in the code, so that seems to be compelling enough for me. 😁

Please, could you fix the tests. Thank you!

kobusjoubert:
Hi @ddnexus, that's awesome! Thank you so much for looking at this pull request 🥳

I just added some tests, runs good on my side.

ddnexus:
Thank you!

PR #685, opened 2024-04-10
Fixed broken aria-label for disabled links in Foundation

daniel-rikowski:
`aria-label="` was missing for disabled links created by the Foundation helpers. Still, the attribute _value_ itself was added, resulting in invalid HTML.

**Before:**

```html
<li class="prev disabled" role="link" aria-disabled="true" Previous>
```

**After:**

```html
<li class="prev disabled" role="link" aria-disabled="true" aria-label="Previous">
```


ddnexus:
@daniel-rikowski good catch and good patch ... missing only the e2e test, but I could add that, because it might not be so simple to setup and, it's just trivial to add.

Thank you!

ddnexus:
Wondering why the e2e test didn't catch the HTML problem when I introduced it. 🤔

daniel-rikowski:
> Wondering why the e2e test didn't catch the HTML problem when I introduced it. 🤔

It looks they are tailored to the broken code, too.

In shapshots.js at line 456 and following:

`     previous=\"\">&lt;</li>`

daniel-rikowski:
> @daniel-rikowski good catch and good patch ... missing only the e2e test, but I could add that, because it might not be so simple to setup and it's just trivial to fix if it's already setup. Please, let me know.

Yes, please. I'm currently on a Windows workstation, I guess it would be even harder...  🙄

ddnexus:
Well... you can add any arbitrary attribute and [HTML-validate](https://html-validate.org) does not complain.
I didn't even find any [rules](https://html-validate.org/rules/index.html) about unknown attributes, so I researched it, and it looks like adding custom attributes to HTML is not invalid HTML! 

The `previous=\"\">&lt;</li>` snapshot normalization of a valid custom attribute happens because `Previous` is treated as a custom empty attribute (which is valid when it omits the `=""` part) on parsing. Then it's probably just normalized as lowercase and dumped with the AST tree in a file as the extended version with the `=""`.

So, no complains about HTML-validate not complaining about perfectly valid HTML 🙃

> I'm currently on a Windows workstation

😱 You have  my sympathy. I will run the tests myself with a deep sense of compassion 😇

Thank you for your help!

daniel-rikowski:
Interesting... I always assumed, custom attributes must be `data-` attributes.  ¯\\_(ツ)_/¯


Glad to be of help 😄

PR #688, opened 2024-04-15
Don't change the state of a 'request' entity using while the '#pagy_url_for' method for calendar pagination

PyrinAndrii:
This PR fixes a bug that I faced trying to implement the calendar pagination.

## Simple explanation:
While doing the calendar pagination I'm getting the next error `expected :page >= 1; got "__pagy_page__"`

## The bug:
I think it was added to the `pagy_url_for` method in this [commit](https://github.com/ddnexus/pagy/commit/8d0b15239b64016c92f3bdc93025390f78c43ef1#diff-51b118db4a7362c98b661b01d662961ed6894e70d1f96ab1988e2ac5802fd423R11-R13), where we assign the `request.Get` object to the `query_params` variable and later change this object in the `pagy_set_query_params` method, so when after it we execute the `pagy_url_for` again the `request.Get` object is different and generates us broken URL.

## How I'm reproducing it in my Rails app
- add `gem 'pagy', '~> 8.1', '>= 8.1.2'` to my Gemfile (Rails project)
- `include Pagy::Frontend` in `ApplicationHelper`
-  `include Pagy::Backend` in my controller
- add `@calendar, @pagy, @meals = pagy_calendar(collection, year: { }, month: { }, day: { })` in my action (e.g. index)
- add several `pagy_nav...` in your view. For example `<%== pagy_nav(@calendar[:year]) %>` & `<%== pagy_nav(@calendar[:month]) %>`

## Check the bug in the [playground](https://ddnexus.github.io/pagy/playground/#4-calendar-app)
The bug is reproducible via Pagy Playground. Look at the images below as the proof:
1. Check generated links before the fix
2. Check generated links after the fix

pagy calendar before fix 

pagy calendar after fix

### DIFF
--- a/gem/lib/pagy/url_helpers.rb
+++ b/gem/lib/pagy/url_helpers.rb
@@ -8,7 +8,7 @@ class Pagy
     # For non-rack environments you can use the standalone extra
     def pagy_url_for(pagy, page, absolute: false, **_)
       vars         = pagy.vars
-      query_params = request.GET
+      query_params = request.GET.clone
       query_params.merge!(vars[:params].transform_keys(&:to_s)) if vars[:params].is_a?(Hash)
       pagy_set_query_params(page, vars, query_params)
       query_params = vars[:params].(query_params) if vars[:params].is_a?(Proc)


ddnexus:
🙏🏻 Thank you for the bug report and fix! Brilliant help from Lviv! 🇺🇦

Sorry for the bug. It originated from 2 things:

1. The `query_params` started from an assignation like `request.GET.merge...`, so it was set as a new object, and the commit changed it, forgetting to `clone` it first
2. We are missing tests for that, because the tests are currently passing 🤦 (and I don't run all the playground apps manually each time)

I am going to release the fix today, and try to add the tests ASAP.
Thank you!

PyrinAndrii:
No problem, happy to help 👍.
Thank you for such a handy gem.

PR #690, opened 2024-04-18
Add page_label translation for ko (closes #592)

### LINKED ISSUE #592: Add ko: "pagination" entry

connie-feng:
I added the Korean translation, which sounds natural in this context if you use the same translation for "page" and "pages".


ddnexus:
Thank you @connie-feng

PR #691, opened 2024-04-19
Docs: count_documents with Mongoid

benkoshy:
Close #670 

(This is a docs change so I don't think git hooks should apply? Pls confirm?)


ddnexus:
Thank you!

PR #695, opened 2024-05-03
Docs: remove irrelevant characters

benkoshy:
Remove the characters - looks like they were part of a mark down panel that got deleted. It is no longer needed.

image


ddnexus:
Thank you!

PR #699, opened 2024-05-16
Typos adjusts

djpremier:
Hi,

These are some small errors that I found while updating my application. If you need it to be in different PRs, I can do that too.

@ddnexus one question I got is how you are running the command to regenerate the Rematch files, looking in its documentation I found the command `rake test TESTOPTS=--rematch-rebuild`, however with this all the files were changed basically in the order in which tests are saved, would this be expected? Or is there some command that forces a certain ordering?

Complete command I used (it can even be mentioned in the contribution doc for those who don't "know" how to run it):
`docker run -it --rm --name pagy-script -v "$PWD":/usr/src/pagy -w /usr/src/pagy ruby:latest /bin/bash -c "bundle install && rake test TESTOPTS=--rematch-rebuild"`

Example of result with divergence:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,7 @@ If you upgrade from version `< 8.0.0` see the following:
 - HTML structure, classes and internal methods have been changed: they may break your views if you used custom stylesheets,
   templates or helper overrides. See the complete changes below if you notice any cosmetic changes or get some exception.
 - The `navs` and `support` extras has been merged into the new [pagy extra](https://ddnexus.github.io/pagy/docs/extras/pagy).
-  Search for `"extra/navs"` and
+  Search for `"extras/navs"` and
   `"extras/support"` and replace with `"extras/pagy"` (remove the duplicate if you used both)
 - The build path for javascript builders has been updated to the canonical paths for gems, and has moved from the `lib` to
   the gem root. Notice that the correct setup in `package json` was still wrongly wrapped in the `gem` dir for 8.0.0-8.0.1, and it
@@ -153,4 +153,4 @@ If you upgrade from version `< 8.0.0` see the following:
 - Docs Improvements
 - Better code issue template
 
-[LEGACY CHANGELOG >>>](CHANGELOG_LEGACY.md) 
+[LEGACY CHANGELOG >>>](CHANGELOG_LEGACY.md)
--- a/CHANGELOG_LEGACY.md
+++ b/CHANGELOG_LEGACY.md
@@ -1,10 +1,10 @@
----          
+---
 icon: versions-24
 layout: page
 visibility: hidden
 ---
 
-# LEGACY CHANGELOG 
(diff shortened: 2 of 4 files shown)


ddnexus:
Thank you @djpremier. Great PR! Code, test and docs fixes! You even removed the trailing spaces at EOL.

_(RubyMine has a built-in option for that but it ignores `.md` files, and I miss a lot of them)_ 🤷

### PR split

- Unless required by code changes, I would prefer to avoid gem updates, that are usually done with npm module updates in their dedicated commits. I believe this PR can avoid it.

- ATM it's better to put the docs fixes in their own PR, so they can be merged directly into `master` and published right away.

- Changes in the `gem` dir (and related test files) must be merged and tested in `dev` so it's easier to have that in a different PR. Notice however that the PR itself should be based on `master`: we will change the base branch before merging.

### Rematch

The `--rematch-rebuild` option rebuilds ALL the rematch files which tests are run, which basically means that it deletes each previous `*.rematch` file and stores whatever are the resulting values from each test file run. 

If you run all files with the rebuild option, you rebuild also the files that didn't fail, thus some value may get reordered inside the new file, despite the actual values are the same. Besides you might unwillingly update unexpected changes of values in `.rematch` files that you didn't check yet and you are not aware of.

A safer way to update the `rematch` files is checking each failure and reconciling one file at the time, either by running the single file with the option, or by manually deleting the `.rematch` file and running the single test file.

djpremier:
Thanks for the feedback @ddnexus 😊.

As for removing spaces, it is automatic when I save a file hehe. As I use VS Code, I adjusted 2 settings in it to avoid conflicts and remove these unwanted "things", they are:
image

### Rematch
Yes, I only left what related to my changes, my question was whether there was anything I should do to maintain order. I will analyze your library when possible, maybe there is something we can adjust so that the result is always the same for the same content.

djpremier:
As for gems, it's a habit of mine to try to keep projects as up to date as possible, to avoid having too many changes to keep track of later, but if you already have your way you can maintain it without any problems. I created the new PR for the code fixes and only uploaded the update necessary for the CI to work.

#700

ddnexus:
> maybe there is something we can adjust so that the result is always the same for the same content

Minitest doesn't run the tests in order, because in the author opinion tests should be atomic and shouldn't need any particular sorting in order to pass. I have to agree.

That means that instead of storing/reading the values from the store file in sync to their executions, we would need to keep the values in some variable. When the test file is done running, sort the values according to the location of the rematch call and write the ordered store file.

We would also have to consider what to do when you just reorder the tests. Should it fail or not? Should we rewrite the store file regardless (for consistency)? Automatically? Writing while we are supposed to read? Should we just warn about it? It looks like we would add quite a lot of complexity.

All that is possible, but I am not sure the benefit of having the ordered store file justifies the overhead and the added maintenance burden, but if you find a better way to do it... you will be my hero! 🙂

Issue #704, opened 2024-05-20
Bug: Pagy 8.4.0 is broken with MS SQL server


erlingur:
### 👀 Before submitting...

- [X] I upgraded to pagy version 8.4.0
- [X] I searched through the [Documentation](https://ddnexus.github.io/pagy/)
- [X] I searched through the [Issues](https://github.com/ddnexus/pagy/issues)
- [X] I searched through the [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)

### 🧐 REQUIREMENTS

- [X] I am providing a VALID code file that confirms the bug
- [X] I am NOT posting any of the USELESS THINGS listed above
- [X] I am aware that this issue will be automatically closed if the code file is missing or INVALID

### 💬 Description

Hi there, I noticed an issue that a very [recent commit](https://github.com/ddnexus/pagy/commit/9a0bffadd1f932a9e085c8922392cd8fef0e6a90) just introduced when using Pagy with a MS SQL Server. Passing a `0` into `.limit()` does not work with MS SQL server. It will give the following error: `The number of rows provided for a FETCH clause must be greater then zero`.

This is because MS SQL server does not support `FETCH NEXT 0 ROWS ONLY`.

Docs: https://learn.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16

Quote:
```
FETCH { FIRST | NEXT } { integer_constant | fetch_row_count_expression } { ROW | ROWS } ONLY
Specifies the number of rows to return after the OFFSET clause has been processed.
The value can be an integer constant or expression that is greater than or equal to one.
```

Here is a Rackup file as requested that demonstrates the issue. I can't spin up a SQL server for you in this file, sorry. You can launch it with a Docker command: `sudo docker run -e "ACCEPT_EULA=Y" -e "MSSQL_SA_PASSWORD=Password.1" -p 1433:1433 --name sql1 --hostname sql1 -d mcr.microsoft.com/mssql/server:2022-latest`

rails.ru:
```
# frozen_string_literal: true

# Starting point to reproduce rails related pagy issues

# DEV USAGE
#    pagy clone rails
#    pagy ./rails.ru

# URL
#    http://0.0.0.0:8000

# HELP
#    pagy -h

# DOC
#    https://ddnexus.github.io/pagy/playground/#2-rails-app

VERSION = '8.4.0'

# Gemfile
require 'bundler/inline'
require 'bundler'
Bundler.configure
[...]

PR #696, opened 2024-05-04
fix slowness on last page (with ransack)

jules-w2:
```ruby
#config/initializers/pagy.rb
Pagy::DEFAULT[:items] = 50
```

On my example, i have 14553 entries and i want the entries of the last page ( page=292)
```ruby
AwsEmail.where("created_at >= ? AND created_at <= ?", "2024-05-01 04:00:00", "2024-06-01 03:59:59").where(:event_type => "Delivery").count
# 14553
```

In my controller
```ruby
def emails
    puts("Start emails")
    t0                      = Time.now
    @search                 = AwsEmail.select(:id).ransack(params[:q])
    @search.sorts           = "id desc" if @search.sorts.empty?
    @search.event_type_eq   = "Delivery"
    @search.created_at_gteq = (Date.today.at_beginning_of_day - 1.day).beginning_of_month.beginning_of_day
    @search.created_at_lteq = (Date.today.at_beginning_of_month.next_month - 1.day).end_of_day
    @pagy, @emails          = pagy(@search.result(:distinct => true))
    puts("emails size => #{@emails.to_a.size}")
    puts(Time.now - t0)
  end
  ```

With the current gem version (LIMIT 50 OFFSET 14550),  it takes 60 seconds
```ruby
Start emails
AwsEmail Count (645.8ms)  SELECT COUNT(*) FROM (SELECT DISTINCT `aws_emails`.`id` FROM `aws_emails` WHERE (`aws_emails`.`event_type` = 'Delivery' AND `aws_emails`.`created_at` >= '2024-05-01 04:00:00' AND `aws_emails`.`created_at` <= '2024-06-01 03:59:59') ORDER BY `aws_emails`.`id` DESC) subquery_for_count
↳ app/controllers/dashboard_controller.rb:48:in `emails'
AwsEmail Load (60024.6ms)  SELECT DISTINCT `aws_emails`.`id` FROM `aws_emails` WHERE (`aws_emails`.`event_type` = 'Delivery' AND `aws_emails`.`created_at` >= '2024-05-01 04:00:00' AND `aws_emails`.`created_at` <= '2024-06-01 03:59:59') ORDER BY `aws_emails`.`id` DESC LIMIT 50 OFFSET 14550
↳ app/controllers/dashboard_controller.rb:49:in `emails'
emails size => 3
60.684757
```

With my pull request (LIMIT 3 OFFSET 14550) it takes 1 seconde
```ruby
Start emails
[...]

PR #706, opened 2024-05-24
Fix: reversion raised in #704

benkoshy:
* Revert [commit 9a0bffadd1f932a9e085c8922392cd8fef0e6a90](https://github.com/ddnexus/pagy/commit/9a0bffadd1f932a9e085c8922392cd8fef0e6a90)

Why?

* [ #704](https://github.com/ddnexus/pagy/issues/704) - ms sql server seems to fail with the above commit.
* Add: troubleshooting docs for the rare case where you have a [slow query on the last page.](https://github.com/ddnexus/pagy/pull/696) - which as the problem commit 9a0bffadd1f932a9e085c8922392cd8fef0e6a90 sought to solve.


benkoshy:
I'll re-do the PR on the proper branch.

PR #716, opened 2024-06-17
Fix pluralization rule link on locale files

rbngzlv:
Quick fix so links on locale files point to the correct location after the reorganizing the gem root dir.

Thank you.


ddnexus:
Thank you1

PR #720, opened 2024-07-08
Docs: fixes link to Arel extra in how to maximize performance docs

AngelGuerra:
Hello! First of all thanks for this awesome gem.

I have found a wrong link in the docs that leads to a 404 page, so this tiny PR points to the correct link.


benkoshy:
Looks good to me!

image

PR #726, opened 2024-07-29
fix: Prefix class names with Pagy module to prevent collisions

LuukvH:
This commit updates the code to prefix certain class names with the Pagy module name. The change is necessary to prevent potential collisions with class names defined in Rails applications.

These adjustments ensure that the Pagy pagination library operates correctly without interfering with other classes that might exist within the application namespace. The revised code now properly isolates Pagy-specific classes, enhancing compatibility and stability

### DIFF
--- a/gem/apps/rails.ru
+++ b/gem/apps/rails.ru
@@ -91,8 +91,9 @@ class Comment < ActiveRecord::Base # :nodoc:
   belongs_to :post
 end # :nodoc:
 
-module Calendar # :nodoc:
-end # :nodoc:
+# Unused model, useful to test overriding conflicts
+module Calendar
+end
 
 # DB seed
 1.upto(11) do |pi|
--- a/gem/apps/rails.ru
+++ b/gem/apps/rails.ru
@@ -63,6 +63,7 @@ end
 require 'pagy/extras/pagy'
 require 'pagy/extras/limit'
 require 'pagy/extras/overflow'
+require 'pagy/extras/headers'
 Pagy::DEFAULT[:limit]    = 10
 Pagy::DEFAULT[:overflow] = :empty_page
 Pagy::DEFAULT.freeze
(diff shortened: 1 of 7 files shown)


ddnexus:
Thank you @LuukvH. Admittedly, the omission is a bit pagy-centric 😁.

LuukvH:
@ddnexus I am not sure why this test is failing. Doesn't seem to relate to my changes? Could you point me in the correct direction, then i can try to fix it.

ddnexus:
Hi @LuukvH , thank you for your persistence.

Please, keep your PR branch rebased on master, and it will take care of the necessary updates that make your branch fail (not your code fault)

ddnexus:
> The change is necessary to prevent potential collisions with class names defined in Rails applications.

@LuukvH I've tried to verify that there would indeed exists a collision, but I couldn't reproduce any of such collisions. Please, could you use any playground app to reproduce it?

ddnexus:
@LuukvH ... Thanks for your example app, but we need a modified playground app to reproduce - as requested in my previous message. I tried to dd a bare `Calendar` model in the `calendar.ru` app, but - as i reported before - it works.

You can see it [here](https://github.com/ddnexus/pagy/compare/master...%23726). Please, modify that app to make it fail, so finding out where is the problem will be trivial.

Thank you.

LuukvH:
@ddnexus i changed the rails playground app as a breaking example. The calendar one is the wrong place to do it, since there Calendar::Unit is actually defined.

ddnexus:
> @ddnexus i changed the rails playground app as a breaking example. The calendar one is the wrong place to do it, since there Calendar::Unit is actually defined.

Where can I find the failing app?

LuukvH:
@ddnexus I added it to this pr. See the rails.ru.

LuukvH:
@ddnexus Let me know if you require anything else.

LuukvH:
@ddnexus Seems a goog improvement to me to absolutize the modules. 
I tested a set of use cases within our source code and this indeed solve the issue we were having.

ddnexus:
Thnk you!

PR #736, opened 2024-09-08
Fix missing `Pagy::` prefix in Searchkick section of config

olieidel:
Hey, thanks so much for Pagy, it looks awesome!

I noticed one minor typo in the `pagy.rb` file when setting it up to work with Searchkick - specifically, one of the `DEFAULT` values which is set is not prefixed with `Pagy::`. Check out the diff :)

Hope it helps and keep up the great work!


ddnexus:
Thank you!

PR #738, opened 2024-09-10
Fix broken links missing "docs" in path

branson-simplethread:
I noticed these two links in the config were broken.


ddnexus:
Thank you!

PR #741, opened 2024-10-03
[FEAT] Add Dzongkha

KinWang-2013:
This PR adds Dzongkha, the national Language of Bhutan 🇧🇹

### DIFF
--- a/gem/lib/pagy/i18n.rb
+++ b/gem/lib/pagy/i18n.rb
@@ -113,6 +113,7 @@ class Pagy
                  hash['zh-CN'] = RULE[:other]
                  hash['zh-HK'] = RULE[:other]
                  hash['zh-TW'] = RULE[:other]
+                 hash['dz']    = RULE[:other]
                end.freeze
     end
 
--- a/gem/locales/dz.yml
+++ b/gem/locales/dz.yml
@@ -1,18 +1,14 @@
-# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/gem/lib/pagy/i18n.rb)
+# :other pluralization (see https://github.com/ddnexus/pagy/blob/master/gem/lib/pagy/i18n.rb)
 en:
   pagy:
     aria_label:
-      nav:
-        one: "ཤོག་ལེབ་"
-        other: "ཤོག་ལེབ་"
+      nav: "ཤོག་ལེབ་"
       prev: "སྔོན་མ་"
       next: "ཤུལ༌མ་"
     prev: "&lt;"
     next: "&gt;"
     gap: "&hellip;"
-    item_name:
-      one: "རྣམ་གྲངས་"
-      other: "རྣམ་གྲངས་"
+    item_name: "རྣམ་གྲངས་"
     info:
       no_items: "%{item_name}གཅིག་ཡང་འཚོལ་མ་འཐོབ།"
       single_page: "%{item_name} %{count} བཀྲམ་སྟོན་འབད་དོ།"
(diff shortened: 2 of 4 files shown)


gem/locales/dz.yml
```diff
@@ -0,0 +1,21 @@
+# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/gem/lib/pagy/i18n.rb)
+en:
```
benkoshy, line 2: should this be a `dz`?

benkoshy:
Hi @KinWang-2013 

Thank you for your PR

- Please confirm that this is a `:one_other` pluralization? I couldn't find `dz` [in the list here](https://github.com/svenfuchs/rails-i18n/tree/master/rails/pluralization) and was wanting to double check to be sure?

gem/locales/dz.yml
KinWang-2013: oh yeah, thanks

KinWang-2013:
> Hi @KinWang-2013
> 
> Thank you for your PR
> 
> * Please confirm that this is a `:one_other` pluralization? I couldn't find `dz` [in the list here](https://github.com/svenfuchs/rails-i18n/tree/master/rails/pluralization?rgh-link-date=2024-10-04T01%3A00%3A25Z) and was wanting to double check to be sure?

yes it is. I will make a PR there as well

gem/lib/pagy/i18n.rb
```diff
@@ -113,6 +113,7 @@ module P11n
                  hash['zh-CN'] = RULE[:other]
                  hash['zh-HK'] = RULE[:other]
                  hash['zh-TW'] = RULE[:other]
+                 hash['dz']    = RULE[:other]
```
ddnexus, line 116: This line is inconsistent with the pluralization indicated in the `gem/locales/dz.yml` and its entries.

gem/locales/dz.yml
```diff
@@ -0,0 +1,21 @@
+# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/gem/lib/pagy/i18n.rb)
```
ddnexus, line 1: Inconsistent with the `:other` pluralization indicated in the `gem/lib/pagy/i18n.rb` entry

gem/lib/pagy/i18n.rb
KinWang-2013: thank you, will update

gem/locales/dz.yml
KinWang-2013: thank you, will update

ddnexus:
> yes it is. I will make a PR there as well

From the entries in the dictionary, it doesn't look that the pluralization rule is `:one_other`, since they show the same translation. It looks more like `:other`. Please, link an official page for the `dz` locale that shows which pluralization rule applies.

Thank you.

KinWang-2013:
> > yes it is. I will make a PR there as well
> 
> From the entries in the dictionary, it doesn't look like the pluralization rule is `:one_other`, since they show the same translation. It looks more like `:other`. Please, link an official page for the `dz` locale that shows which pluralization rule applies, and fix either the rule or the entries.
> 
> Thank you.

sorry, got a bit confused, will update, I don't think there are any official page for dz locale.

ddnexus:
Maybe we can link https://www.unicode.org/cldr/charts/45/supplemental/language_plural_rules.html at the top of the `dz` dictionary file.

ddnexus:
Thank you

PR #743, opened 2024-10-04
Docs: add link to unicode pluralization rules

benkoshy:
Fix the problem issue raised here:  https://github.com/ddnexus/pagy/pull/741 where we could not easily identify the pluralisation.

Hope that helps.


ddnexus:
Thank you!

PR #745, opened 2024-10-16
Update `Pagy:: Backend` comment

leonvogt:
Hi 👋

I've noticed that the overriding  of `pagy_get_vars` stopped working after the 9.0.0 upgrade.     
It seems this method was removed during [this refactoring](https://github.com/ddnexus/pagy/commit/26b8117fe547f9ec5dfd75b248fbe31910d77fc1#diff-e6c288388e84855ee6d63cfbcc7cb692986aa68e7683af84ad98eed4cb95eb93R55).     

In my use case, I was able to switch to overriding `pagy_get_count` instead.

I've updated the comment to reflect this change.

### DIFF
--- a/gem/lib/pagy/backend.rb
+++ b/gem/lib/pagy/backend.rb
@@ -1,5 +1,6 @@
 # See Pagy::Backend API documentation: https://ddnexus.github.io/pagy/docs/api/backend
 # frozen_string_literal: true
+# You can override any of the `pagy_*` methods in your controller.
 
 class Pagy
   # Define a few generic methods to paginate a collection out of the box,
--- a/gem/lib/pagy/backend.rb
+++ b/gem/lib/pagy/backend.rb
@@ -3,7 +3,7 @@
 
 class Pagy
   # Define a few generic methods to paginate a collection out of the box,
-  # or any collection by overriding pagy_get_items and/or pagy_get_vars in your controller
+  # or any collection by overriding any of the `pagy_get_` methods in your controller.
   # See also the extras if you need specialized methods to paginate Arrays or other collections
   module Backend
     private
--- a/gem/lib/pagy/backend.rb
+++ b/gem/lib/pagy/backend.rb
@@ -3,7 +3,7 @@
 
 class Pagy
   # Define a few generic methods to paginate a collection out of the box,
-  # or any collection by overriding any of the `pagy_get_` methods in your controller.
+  # or any collection by overriding any of the `pagy_get_*` methods in your controller.
   # See also the extras if you need specialized methods to paginate Arrays or other collections
   module Backend
     private
--- a/gem/lib/pagy/extras/arel.rb
+++ b/gem/lib/pagy/extras/arel.rb
@@ -1,5 +1,6 @@
 # See the Pagy documentation: https://ddnexus.github.io/pagy/docs/extras/arel
 # frozen_string_literal: true
+# You can override any of the `pagy_*` methods in your controller.
 
 class Pagy # :nodoc:
   # Better performance of grouped ActiveRecord collections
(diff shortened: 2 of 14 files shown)


gem/lib/pagy/backend.rb
```diff
@@ -3,7 +3,7 @@
 
 class Pagy
   # Define a few generic methods to paginate a collection out of the box,
-  # or any collection by overriding pagy_get_items and/or pagy_get_vars in your controller
+  # or any collection by overriding pagy_get_items and/or pagy_get_count in your controller
```
benkoshy, line 6: Nicely spotted!

```diff
-  pagy_get_vars
+  pagy_get_count
```

Any of the `pagy_get_` methods may be overridden (for whatever reason): should we alert users to this possibility rather than confine the patch to just `pagy_get_items` and `pagy_get_count`?

ddnexus: > Any of the `pagy_get_*` methods...

👍🏼 @benkoshy since it's a code comment... yes, a general comment may be more appropriate.

We should consider also adapting it to the backend extras.

ddnexus:
This PR requires a rebase on master in order to pass the tests.

gem/lib/pagy/backend.rb
leonvogt: @benkoshy Ah yes fair enough. I wasn't sure so I went for the least changes.     
Ive updated the comment to:
```ruby
# Define a few generic methods to paginate a collection out of the box,
# or any collection by overriding any of the `pagy_get_` methods in your controller.
# See also the extras if you need specialized methods to paginate Arrays or other collections
```

Do you think that's better? Or do you have some better suggestions?

benkoshy: > Do you think that's better? Or do you have some better suggestions?

Looks great!

BTW: if you want bonus marks - consider adapting a similar comment in the "extras". i.e. to alert to the possibility that those methods can be overridden.  Or I can do it for you and credit you with the PR.

leonvogt: > consider adapting a similar comment in the "extras".

Hm, are you talking about the [Pagy::PagyExtra](https://github.com/ddnexus/pagy/blob/master/gem/lib/pagy/extras/pagy.rb)?    
I've never overridden those and I'm not sure if I can provide a clear comment instruction on how to override them.    
Seems like you need make some [extra steps to override frontend methods.](https://benkoshy.github.io/2020/02/01/overriding-pagy-methods.html#user-content-b-pagy-front-end-methods)

I would be happy if someone else with actual experience can provide a comment there.

benkoshy: > Hm, are you talking about the [Pagy::PagyExtra](https://github.com/ddnexus/pagy/blob/master/gem/lib/pagy/extras/pagy.rb)?

- Pagy has some "[extras](https://github.com/ddnexus/pagy/tree/master/gem/lib/pagy/extras)" with methods that be overriden. A comment similar to what you suggested above would suffice?

benkoshy: I have updated with a note suggesting that backend methods can be overridden and have credited Leon Vogt

if you are satisfied i can squate / rebase on master?

benkoshy: here is the full branch with missing comments (sorry I forgot to paste it): https://github.com/ddnexus/pagy/commits/update-backend-comment/ 

- https://github.com/ddnexus/pagy/commit/81b781621fe7b2cd3288e242266fba8448c30856
- https://github.com/ddnexus/pagy/commit/3fbf56be68f98b8a31049b869ecb186cf4a1e402

Perhaps cherry pick those comments, remove my name, add yours, and you'll be winning.

leonvogt: Awesome thank you! I've cherry picked the two commits. And I will for sure not remove your name! You have done all the work! 😂
Do you want me to squash them or are you doing that while merging?

ddnexus:
Thank you!

PR #765, opened 2025-03-07
Fix method visibility in JsonApiExtra

yenshirak:
If you load the Jsonapi extra and include `Pagy::Backend` in a controller then `pagy_url_for` and `pagy_set_query_params` will be added as public methods. Only the controller methods that are mapped to routes should be public.


ddnexus:
Thank you!
@yenshirak FYI, the upcoming v10 will not have any scattered around methods like this anymore.

PR #771, opened 2025-03-15
Docs: add transparency to images

benkoshy:
### What is this PR?

- It adds a "transparency channel" to many of the images in the docs.
- And makes certain portions of the image transparent, where it makes sense to do so. In same cases, adding a transparency to all areas would not make sense, especially with dark backgrounds.

Please review and LMK if any changes are required?


benkoshy:
I've made the PR using the wrong images on the wrong branch.

🤦🤦

PR #772, opened 2025-03-16
Docs: add transparency to images

benkoshy:
### What is this?

We wanna make the pics better looking on all backgrounds: whether dark mode or otherwise. Otherwise you will have white backgrounds glaring at you while in dark mode. It's workable, but not ideal.

To make an image transparent, we add what's called an "alpha channel" to all images, and mark certain areas as transparent where required. This is done using image editing software. And we must save it as a .png.

This PR adds that transparency to all pagination images (charts have been ommitted).

Any issues pls LMK.


ddnexus:
Thank you!

Issue #773, opened 2025-03-21
Bug: `jsonapi` and `trim` extras break Pagy nav helpers when used simultaneously


marcus-deans:
### 👀 Before submitting...

- [x] I upgraded to pagy version 9.3.4
- [x] I searched through the [Documentation](https://ddnexus.github.io/pagy/)
- [x] I searched through the [Issues](https://github.com/ddnexus/pagy/issues)
- [x] I searched through the [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)
- [x] I asked [Pagy Trained AI](https://gurubase.io/g/pagy) - Warning - The information may not be 100% accurate, but it serves as a good starting point for further investigation.

### 🧐 REQUIREMENTS

- [x] I am providing a VALID code file that confirms the bug
- [x] I am NOT posting any of the USELESS THINGS listed above
- [x] I am aware that this issue will be automatically closed if the code file is missing or INVALID

### 💬 Description

Using default configuration for `jsonapi` and `trim` at the same time  breaks Pagy nav helpers:

I have attached an edited `demo.ru` file, but for convenience, the only changes I made was adding
```diff
+ require 'pagy/extras/jsonapi'
+ # Pagy::DEFAULT[:trim_extra] = false # opt-in trim
- Pagy::DEFAULT[:trim_extra] = false # opt-in trim
```

Thus, the default configuration for both `jsonapi` and `trim` is implemented. 

The demo app works fine for the initial load. It then breaks for any subsequent pages when trying to load the prior page's URL. This is observable below

https://github.com/user-attachments/assets/c2d943ea-fdd5-4b7c-9b35-49fa5e2a680a

marcus-deans:
I understand the report includes these guidelines:

> 🚩 Avoid posting USELESS THINGS like:
> * Code snippets / Log transcripts
> * Descriptions of errors / behaviors / expectations
> * Rationalizations about WHY you THINK your problem is a pagy bug

So did not include this in my actual report. However, I believe I can directly identify the problem and would like to save your debugging time, if helpful.

The issue is that `jsonapi` automatically overrides the `pagy_set_query_params` method, such that page params throughout the application become nested

```ruby
module UrlHelperOverride
      # Override UrlHelper method
      def pagy_set_query_params(page, vars, query_params)
        return super unless vars[:jsonapi]

        query_params['page'] ||= {}
        query_params['page'][vars[:page_param].to_s]  = page
[...]

PR #774, opened 2025-03-21
Support JSON:API Parsing in `Trim`

marcus-deans:
This is a V1 draft just to show the concept I alluded to in #773.

I am not sure about the consistent availability of `pagy.vars[:jsonapi]` across requests. However, I hope this is somewhat helpful as an initial parsing testing implementation, in case this avenue is pursued

*I'm not sure about preferred comment/test structure either*

As a demo of the changes, here is the demo test app setup as described in #773, showing the implementation working

https://github.com/user-attachments/assets/f7483827-42be-4bae-83e4-e2b426e081d3

### DIFF
--- a/test/pagy/extras/trim_test.rb
+++ b/test/pagy/extras/trim_test.rb
@@ -2,7 +2,7 @@
 
 require_relative '../../test_helper'
 require 'pagy/extras/trim'
-
+require 'pagy/extras/jsonapi'
 require_relative '../../mock_helpers/app'
 
 describe 'pagy/extras/trim' do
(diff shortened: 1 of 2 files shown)


test/pagy/extras/trim_test.rb
```diff
@@ -28,14 +28,48 @@
...
-        pagy = Pagy.new(count: 1000, page: page, trim_extra: false)
+        pagy = Pagy.new(count: 1000, page: page, jsonapi: false, trim_extra: false)
         a    = app.pagy_anchor(pagy)
         _(a.(page)).must_include(%(<a href="/foo#{not_trimmed}"))
       end
     end
+
+    describe 'when jsonapi is enabled' do
+      it 'returns trimmed or not trimmed links' do
+        [ # first page
```
marcus-deans, line 43: I found that repeating all of the expectations kept things explicit, and avoided any unintentional false positives that might occur through transforming the pre-existing examples above

ddnexus:
Thank you for your PR. Not sure about what to do given my comment in #773. Thoughts?

PR #780, opened 2025-05-29
fix column names quotation in keyset

t0ch1k:
This PR should fix the problem with column names quotes in MySQL introduced by [this commit](https://github.com/ddnexus/pagy/commit/8b7833934d5ba1d9e1767dd5649b1a322bb83072). 

Demo of the problem:
```
irb(main):002> scope = Service.order(id: :desc)
irb(main):003> page = Pagy::Keyset.new(scope, limit: 3000)
irb(main):004> page.records.count
  Service Load (25.3ms)  SELECT `services`.* FROM `services` ORDER BY `services`.`id` DESC LIMIT 3001
irb(main):006> page = Pagy::Keyset.new(scope, limit: 3000, page: page.next)
  Service Load (2.9ms)  SELECT `services`.* FROM `services` /* loading for pp */ ORDER BY `services`.`id` DESC LIMIT 11
=>
#<Pagy::Keyset::ActiveRecord:0x0000000109b270f0
...
irb(main):007> page.records.count
  Service Load (2.3ms)  SELECT `services`.* FROM `services` WHERE (( "services"."id" < '7002' )) ORDER BY `services`.`id` DESC LIMIT 3001
(irb):7:in `<main>': Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '."id" < '7002' )) ORDER BY `services`.`id` DESC LIMIT 3001' at line 1 (ActiveRecord::StatementInvalid)
/Users/alexeyt/.rbenv/versions/3.1.6/lib/ruby/gems/3.1.0/gems/mysql2-0.5.6/lib/mysql2/client.rb:151:in `_query': You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '."id" < '7002' )) ORDER BY `services`.`id` DESC LIMIT 3001' at line 1 (Mysql2::Error)
```

PR #783, opened 2025-06-25
Fix keyset name quoting for MySQL

benjaminwood:
This change ensures that the keyset column names are properly quoted when using MySQL.


t0ch1k:
Doesn't it duplicate my PR #780 ?

benjaminwood:
Oh hey @t0ch1k you're right! Thanks for pointing it out, and for your contribution to fix this issue. Closing my PR here now.

Looking forward to yours being merged 😄

PR #791, opened 2025-08-15
Docs: add pagy wand demo video links

benkoshy:
* Add links to the pagy wand video demo
* The links are styled as "images" which link to a youtube video.

image

Cons:

- there is too much white noise in the video. I will see if I can upload a better audio version without that noise. Perhaps with a better mic too.


ddnexus:
Thank you Ben

PR #796, opened 2025-09-05
Docs: add pagy options exception handling

benkoshy:
Sometimes bots spam urls, causing exceptions to be raised, which may trigger emails to be sent, perhaps unnecessarily.

This documents a solution, and is pretty much the same as with handling `RangeErrror` exceptions. 

Full credit to @excid3 for this PR.

### DIFF
--- a/docs/guides/how-to.md
+++ b/docs/guides/how-to.md
@@ -369,6 +369,18 @@ def redirect_to_last_page(exception)
 end
 ```
 
+#### Manage `Pagy::OptionError` Exceptions
+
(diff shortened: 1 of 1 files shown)


ddnexus:
It would probably be a bit better to check `e.option == :page` before assuming that it's actually a bad bot affecting the `:page` and redirect unconditionally.

Skipping that check would mask all other legit errors.

excid3:
I actually do want to handle _any_ `OptionError` in case bots were messing with other params, but it might be good to note how to handle specific cases like that too.

ddnexus:
Thank you @benkoshy @excid3

PR #797, opened 2025-09-11
Support queries with SQL order

seanpdoyle:
When passed an `ActiveRecord::Relation` that has been ordered with `String` or `Arel::Nodes::SqlLiteral` instance, rely on the SQL `ORDER` clause.

### DIFF
--- a/test/pagy/keyset_test.rb
+++ b/test/pagy/keyset_test.rb
@@ -86,6 +86,21 @@ require 'pagy/keyset'
 
         _(pagy.instance_variable_get(:@keyset)).must_equal({animal: :desc, :id => :asc})
       end
+      if model == Pet
+        it 'extracts the keyset from the set order (SQL Literal)' do
+          set = model.order(Arel.sql("animals DESC, id ASC"))
+          pagy = Pagy::Keyset.new(set)
+
+          _(pagy.instance_variable_get(:@keyset)).must_equal(Arel.sql("animals DESC, id ASC"))
+        end
+
+        it 'extracts the keyset from the set order (String)' do
+          set = model.order("animals DESC, id ASC")
+          pagy = Pagy::Keyset.new(set)
+
+          _(pagy.instance_variable_get(:@keyset)).must_equal("animals DESC, id ASC")
+        end
+      end
       if model == PetSequel
         it 'raises TypeError for unknown order type' do
           _ { Pagy::Keyset.new(model.order(id: :desc)) }.must_raise TypeError


seanpdoyle:
Unfortunately, this PR in its current form only introduces test coverage. That coverage fails for `String` and `Arel::Nodes::SqlLiteral` instances.

Our application is attempting to paginate an `ActiveRecord::Relation` constructed by the [pg_search](https://github.com/Casecommons/pg_search/) gem. Under the hood, that gem's query orders by an SQL literal, resulting in the following error:

```
NoMethodError: undefined method `value' for an instance of Arel::Nodes::SqlLiteral
```

The current Keyset implementation expects that the ActiveRecord::Relation has only called `.order` with a `Hash` or keyword arguments.

I'm happy to attempt to submit an implementation that passes the test, but I worry that I'd need to make a large amount of changes to do so. I defer to this gem's maintainers if there's a more straightforward solution.

ddnexus:
@seanpdoyle thank you for the PR.

The most straightforward way to assign the right order, in cases where the default extraction wouldn't work, is to override the `@keyset` variable. In the current master (which will be soon the legacy pagy) you would have to override the `extract_keyset` method 🤷.

In the upcoming v43 (currently rc4), you can pass the `:keyset` option and it will skip the default extraction and just use it https://github.com/ddnexus/pagy/blob/2b2a0467de780c884af8be87a05eef2b6aeb8ee3/gem/lib/pagy/classes/keyset/keyset.rb#L37

That way you can manually set it at will for your specific set, and pagy wouldn't actually need to implement it.

Most app shouldn't need any special handling of it, so we can keep pagy slim, still allowing easy customization. That should be added in the docs (probably an entry in the "How to" page). 

However in case the handling of less common cases could be concisely implemented, we may consider including it.

Is that an acceptable policy?

Splines:
Would love to see support for this. We have a complex [ordering](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/models/lecture.rb#L149-L152) and would like to still use `keyset` navigation with it.

```rb
def self.default_search_order
   # NOTE: This requires the query to join :course and :term
   Arel.sql("terms.year DESC, terms.season DESC, LOWER(unaccent(courses.title)) ASC")
end
```

ddnexus:
@Splines thanks for the perfect example. That's exactly why it's not supported and not even supportable.

That ordering is not even indexable because it spans two separate tables, and [PostgreSQL indexes](https://www.postgresql.org/docs/current/indexes.html?hl=en-US) are restricted to a single table.

Without an efficient index, the benefits of keyset pagination are gone.

But even if a proper index was possible, how can pagy reliably and automatically extract a `@keyset` to use in its [compose_predicate method](https://github.com/ddnexus/pagy/blob/b8575c56dcc7a232e2b0c1f2073bba4420b00b68/gem/lib/pagy/classes/keyset/keyset.rb#L113) from that text?

And if you know what the `@keyset` would be, then why would you need to pass through the huge amount of code and overhead that pagy would require... just to find out a simple value that you already know? In that case you could just pass the `keyset: your_known_keyset` [(already supported) option](https://ddnexus.github.io/pagy/toolbox/paginators/keyset/#options) and you (and pagy) would be done and happy to use keyset.

HTH

Splines:
Thanks for your insights!

> In that case you could just pass the `keyset:`

I'd love to do that. However, for the given example, when I pass the keyset as

```rb
keyset: { year: :desc, season: :desc, title: :asc }
```

we get the error `undefined method 'year' for an instance of Lecture` since the ordering depends on a prior join of the `lecture` table with the `course` and `term` table. I'm not sure what a solution to that would be.

---

To give some more context for this:

- Search results are computed [here](https://github.com/MaMpf-HD/mampf/blob/1d8a6ad644a2a3fa667863524de56c2a6e6869f8/app/search/searchers/controller_searcher.rb#L39-L43), which uses [`ModelSearcher.search()`](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/search/searchers/model_searcher.rb#L12-L24), which uses [`SearchSorter.sort()`](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/search/sorters/search_sorter.rb#L10-L14), which [applies the default order](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/search/sorters/search_sorter.rb#L44-L56) that it got from [`order_expression`](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/search/sorters/search_sorter.rb#L55), which it memoizes from [here](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/search/sorters/search_sorter.rb#L29-L32), i.e. for example for the lecture [here](https://github.com/MaMpf-HD/mampf/blob/3b7ee0508df8c77ce1689a38ce9994cf3ddac297/app/models/lecture.rb#L149-L156).
- Pagy is then called [here](https://github.com/MaMpf-HD/mampf/blob/1d8a6ad644a2a3fa667863524de56c2a6e6869f8/app/search/searchers/controller_searcher.rb#L53).

Splines:
Unfortunately, I couldn't get this scenario to work with Pagy, so after a lot of back and forth with various AIs, this is now a custom solution that is working fine. Feel free to take inspiration should this be something Pagy wants to include, but I totally understand if this is out of scope.

If you have any better ideas, I'd be happy to hear about them, especially since I've never had too many points of contact with pagination in the past.

### 🎈 Custom solution for keyset pagination with SQL order

In our model, we have defined a custom search order

```rb
def self.default_search_order
  # NOTE: This requires the query to join :course and :term
  Arel.sql("terms.year DESC, terms.season DESC, LOWER(unaccent(courses.title)) ASC")
end
```

that is used as described [in my comment above](https://github.com/ddnexus/pagy/pull/797#issuecomment-3689411536). With the new solution, we can now use it as follows in the [`ControllerSearcher`](https://github.com/MaMpf-HD/mampf/blob/4473b29582018fe3117436abb9f04c1d0c386543/app/search/searchers/controller_searcher.rb#L47-L53):
```rb
keyset = Search::Pagination::OrderParser.keyset_from(model_class.default_search_order)
Search::Pagination::KeysetPager.paginate(
  set: search_results,
  keyset: keyset,
  limit: default_per_page,
  page: config.params[:page]
)
```

with the custom [`KeysetPager`](https://github.com/MaMpf-HD/mampf/blob/4473b29582018fe3117436abb9f04c1d0c386543/app/search/pagination/keyset_pager.rb#L1-L75) and the [`OrderParser`](https://github.com/MaMpf-HD/mampf/blob/4473b29582018fe3117436abb9f04c1d0c386543/app/search/pagination/order_parser.rb#L1-L50). The `OrderParser` is also used in our `SearchSorter` [here](https://github.com/MaMpf-HD/mampf/blob/4473b29582018fe3117436abb9f04c1d0c386543/app/search/sorters/search_sorter.rb#L43-L72).


### 🎈 Demo

From the branch [`feature/lecture-browse`](https://github.com/MaMpf-HD/mampf/tree/feature/lecture-browse). In the video, the next page is automatically retrieved when the user scrolls further down.

https://github.com/user-attachments/assets/e355b456-37ff-46ac-83dd-69fa0b170a6c

ddnexus:
@Splines I am so sorry for not having been more explicit in my previous message... I just realized that I wasn't clear enough.

My point was that your 
> ordering is not even indexable because it spans two separate tables, and PostgreSQL indexes are restricted to a single table.

The implicit consequence of that is that without a proper index the keyset pagination has only limitations and no advantages, hence you are probably better off using an OFFSET paginator.

All the code that you had to write may make it work, but it's like making a F1 car work on diesel fuel, expecting to win the GP Championship.

> If you have any better ideas, I'd be happy to hear about them

Yes, do not use keyset without a proper index. If you can't have one, then save all the complications added to your code and use one of :offset ,  :countish, :countless paginators.

Or if it is somehow possible in your case, create a single table where the keyset is indexable (e.g. duplicate/consolidate the data from joined table into one).

Or change the DB. In certain DBs, you can create a an indexed/materialized view or join indexes that behave like having a single index that can span multiple tables. 

That's a sophisticated topic that you have to research in the details of the DB implementation, however the general rule is that in order to use keyset conveniently, you have to optimize your data for indexing.

HTH

Splines:
Thanks a lot for these clarifications. 

> All the code that you had to write may make it work, but it's like making a F1 car work on diesel fuel, expecting to win the GP Championship.

Haha ;)

I removed my own code now and just went with `:countless` and it works perfectly fine for our infinite scroll scenario. I probably hopped too quick onto the keyset train as I read this in the [docs for `:countless`](https://ddnexus.github.io/pagy/toolbox/paginators/countless/):

> Consider using the `:keynav_js` paginator when possible!
> The [:keynav_js](https://ddnexus.github.io/pagy/toolbox/paginators/keynav_js/) offers identical UI features but utilizes the faster keyset pagination.

But of course, for that premise to be true, you need a proper index first, which I missed not knowing how keyset pagination works internally. Now I get it, but maybe it'd be worth to add a caveat to the [`:keyset`](https://ddnexus.github.io/pagy/toolbox/paginators/keyset/) and [`:keynav_js`](https://ddnexus.github.io/pagy/toolbox/paginators/keynav_js/) docs  for other users.

ddnexus:
I thought we wrote that in the [constraints](https://ddnexus.github.io/pagy/toolbox/paginators/keyset/#constraints) and if you're in keynav_js there's is a call-out like this:

> [!Warning]
> **This documentation integrates the :keyset documentation.**
> It's easier to understand if you familiarize with the :keyset docs.

It's probably easy to miss. I will see what we can do. Thank you.

PR #799, opened 2025-09-13
Docs: add note and issue template - where to make PRs

benkoshy:
This PR mirrors #798 made on the `master-pre` branch.

It needs to be replicated on both branches.


ddnexus:
Implemented by cherry-picking the #798 from dev
Thank you @benkoshy

Issue #605, opened 2024-01-10
Add tr: "pagination" entry


ddnexus:

benkoshy:
I believe we can close this because Turkish uses the "other" translation.

ddnexus:
Dobee have the `tr` translation for "pages"?

benkoshy:
> Dob we have the `tr` translation for "pages"?

We do not have the translation. I made the fatal assumption that "pages" would be the translation. Which means I will re-open a few issues.

Cosmo:
I hope a comment is fine -- if not, please let me know so I can create a PR.

```yml
tr:
  pagy:
    p11n: 'OneOther'
    aria_label:
      nav:
        one: "Sayfa"
        other: "Sayfalar"
      previous: "Önceki"
      next: "Sonraki"
    previous: "&lt;"
    next: "&gt;"
    gap: "&hellip;"
    item_name:
      one: "kayıt"
      other: "kayıtlar"
    info_tag:
      no_count: "Sayfa %{page} / %{pages}"
      no_items: "Hiç %{item_name} bulunamadı"
      single_page: "%{count} %{item_name} gösteriliyor"
      multiple_pages: "%{count} toplamda %{item_name} %{from}-%{to} arası gösteriliyor"
    input_nav_js: "Sayfa %{page_input} / %{pages}"
    limit_tag_js: "Sayfa başına %{limit_input} %{item_name} göster"
```

ddnexus:
A PR would keep better track of things, also giving you the credit...

Cosmo:
This part made me post a comment instead 😅

https://github.com/ddnexus/pagy/blob/40e9a26f41193e70f92b67cfceee14462f14e4e9/gem/locales/tr.yml#L7-L8

PR incoming

ddnexus:
Yeah... that comment is there to try to get contributions also from lazy people... but you offered a PR 🤷🤭.

Thank you for your help! 🙏

Cosmo:
Haha, love the approach :D

Thanks for the Gem btw!.

PR #824, opened 2025-11-12
Update tr.yml

Cosmo:
This adds singular and plural translations for "Pages", "Item" and improved translation of page summary messages in `info_tag.single_page` and `info_tag.multiple_pages` for more natural Turkish phrasing.
Also I revised `limit_tag_js` to clarify that the limit applies per page.

See: #605


ddnexus:
Thank you!

PR #828, opened 2025-11-17
Allow passing nested `page_key` param

krzyzak:
This PR enables support for nested page keys in paginated requests, allowing multiple paginated resources to coexist on the same page with distinct pagination state.

When rendering multiple paginated collections on a single page, I needed a way to namespace pagination parameters to avoid conflicts. For example, on a dashboard displaying both paginated posts and paginated users, we need separate pagination controls for each collection (eg: `{ posts: { author_id: 123, page: 4 }, users: { page: 5 } }`)

This PR allows passing `page_key` as an array (e.g., `["posts", "page"]`) to access nested pagination parameters. This enables the paginator to read from `params[:posts][:page]` instead of the default `params[:page]`.

### DIFF
--- a/gem/lib/pagy/classes/request.rb
+++ b/gem/lib/pagy/classes/request.rb
@@ -20,7 +20,7 @@ class Pagy
 
     def resolve_page(options, force_integer: true)
       page_key = options[:page_key] || DEFAULT[:page_key]
-      page     = @jsonapi ? @params['page'][page_key] : @params[page_key]
+      page     = @jsonapi ? @params['page'][page_key] : @params.dig(*page_key)
       page     = nil if page == ''   # fix for app-generated queries like ?page=
       force_integer ? (page || 1).to_i : page
     end


ddnexus:
@krzyzak Thank you for the idea.

Traditionally pagy doesn't nest params (jsonapi is the exception... for compliance). If you want them in the same url, you prepend a prefix, which keeps the params flat (and keeps the pagy code simple). For example, we use prefixes in the calendar pagination.

Indeed pagy needs to extract page and limit from the params, but it also needs to generate URLs consistent with the structure. 

BTW: your change seems to miss the URL generation of the array keys that you suggest to use, beside missing the same changes for the limit.

However, your PR made me think that since pagy already handle nested page and limit params for the jsonapi  automatically, prefixing params doesn't actually make the code simpler anymore. 

Tech note: We could generalize the jsonapi code to handle a new 'root_key' param instead, and the jsonapi case will internally become simply the case of the 'root_key' set to 'page'. The change should be very simple and would allow read/write nested param in any root_key. It could also be applied to the calendar code, which could avoid prefixing, making it more readable.

What do you think?

krzyzak:
Hey @ddnexus, thank you for quick response!

> Indeed pagy needs to extract page and limit from the params, but it also needs to generate URLs consistent with the structure.

> BTW: your change seems to miss the URL generation of the array keys that you suggest to use, beside missing the same changes for the limit.

You're right. I posted this PR as a PoC to check if you like the idea and if there's a chance to get it merged. I indeed patched the url generation method to support my case in my project.

> Tech note: We could generalize the jsonapi code to handle a new 'root_key' param instead, and the jsonapi case will internally become simply the case of the 'root_key' set to 'page'. The change should be very simple and would allow read/write nested param in any root_key. It could also be applied to the calendar code, which could avoid prefixing, making it more readable.

Well, while I don't know much about the calendar code, I agree that the jsonapi case would be as simple as a `root_key` set to `page`. I'm just unsure if I have enough context and know the code well enough to deliver those changes

ddnexus:
> I'm just unsure if I have enough context and know the code well enough to deliver those changes

Don't worry, I was just asking for your feedback... I already implemented the changes. It works with the `:jsonapi` as before, and with any `:root_key`.

You can try the `dev` branch and hopefully confirm that it works.

n-rodriguez:
> You can try the `dev` branch and hopefully confirm that it works.

@ddnexus it works! thank you!

ddnexus:
Implemented in [43.1.0](https://github.com/ddnexus/pagy/releases/tag/43.1.0)

PR #836, opened 2025-11-25
Fix root_key params deletion bug

### LINKED ISSUE #835: Bug: compose_page_url deletes all params under root_key, not just pagination keys
### 👀 Before submitting...

- [x] I upgraded to pagy version 43.1.4
- [x] I asked [Pagy AI](https://gurubase.io/g/pagy) (Good starting point for further investigation).
- [x] I searched through the [Documentation](https://ddnexus.github.io/pagy/), [Issues](https://github.com/ddnexus/pagy/issues), and [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)

### 🧐 REQUIREMENTS

- [x] I am providing a VALID code file that confirms the bug
- [x] I am NOT posting any of the USELESS THINGS listed above
- [x] I am aware that this issue will be automatically closed if the code file is missing or INVALID

### Description
When using `root_key`, the `compose_page_url` method deletes the entire nested hash instead of only removing pagination-specific keys. This causes all other params
 nested under `root_key` to be lost.

### Current Behavior
Starting params: `{ "users" => { "page" => "1", "q" => "search_term" } }`
After compose_page_url with root_key: `users`
Result: `{ "users" => { "page" => "2" } }`
Lost: `users[q]` parameter

### Expected Behavior

The `users[q]` parameter should be preserved in the URL.

### Root Cause

Two issues in `gem/lib/pagy/modules/abilities/linkable.rb`:

Line 44:
```ruby
params.delete(root_key || page_key)
```
Deletes the entire root_key hash instead of just page_key within it.

Line 49:
```ruby
params.merge!(root_key ? { root_key => factors } : factors)
```
Replaces the entire root_key hash with only factors, losing any other nested params.

### Failing Test

I've added a failing test that demonstrates this bug:
https://github.com/antoninbena/pagy/commit/9128ffcd034cd2a3820d56cbec160412db01b0fc

### Environment

- Pagy version: 43.1.4

antoninbena:
Fixes #835

## Summary
Fixed `compose_page_url` to preserve nested params under `root_key` instead of deleting the entire hash.

## Changes
- Only delete `page_key` from within `root_key` hash, not the entire hash
- Properly dup nested hash to avoid mutating original request params (clone is shallow)
- Merge factors into existing nested hash instead of replacing it
- Add test to verify nested params are preserved

## Testing
All tests pass including the new test case that demonstrates the fix.

### DIFF
--- a/gem/lib/pagy/modules/abilities/linkable.rb
+++ b/gem/lib/pagy/modules/abilities/linkable.rb
@@ -41,12 +41,23 @@ class Pagy
         @options.merge(options)
                 .values_at(:root_key, :page_key, :limit_key, :limit, :client_max_limit, :querify, :absolute, :path, :fragment)
       params = @request.params.clone(freeze: false)
-      params.delete(root_key || page_key)
+      if root_key
+        params[root_key] = params[root_key]&.dup || {} # clone is shallow, need to dup nested hash
+        params[root_key].delete(page_key)
+      else
+        params.delete(page_key)
+      end
       factors = {}.tap do |h|
                   h[page_key]  = countless? ? RawQueryValue.new(compose_page_param(page)) : page
                   h[limit_key] = limit_token || limit if client_max_limit
                 end.compact # No empty params
-      params.merge!(root_key ? { root_key => factors } : factors) if factors.size.positive?
+      if factors.size.positive?
+        if root_key
+          params[root_key] = factors.merge(params[root_key])
+        else
+          params.merge!(factors)
+        end
+      end
       querify&.(params) # Must modify the params: the returned value is ignored
       query_string = QueryUtils.build_nested_query(params)
       query_string = "?#{query_string}" unless query_string.empty?


ddnexus:
About the failing tests: the change in the `compose_page_url` modified the order of the query string, so I added a different assertion/expectation in the tests (`assert_url_equal`/`must equal_url`), which ignores the order, and passes the old tests, making testing url strings a lot easier. See https://github.com/ddnexus/pagy/commit/ef197a05632795195f9f0b6ed1476f0999de0cfe

ddnexus:
Thank you for your help!

antoninbena:
Thanks for your time!
I didn't want to rewrite all of those tests, so my solution preserved the original param order. But this is much cleaner, so thanks for taking your time and doing it correctly.

antoninbena:
Actually, I found a new issue with this. It's only relevant to Rails, so I'm not sure if it belongs here.

## Problem

Pagination with `root_key` now breaks in Rails when `params[:root_key]` are empty.

**Demo:** Here's a [demo app](https://gist.github.com/antoninbena/e0926c723fb1d3dd63e025a4aae4389d) that shows the pagination not working.

## Investigation

I did some investigation and found that the problem is in:
```ruby
(root_key ? params[root_key] ||= {} : params).tap do |h|
```

It works with normal hashes, but Rails uses `ActiveSupport::HashWithIndifferentAccess` for the params. In this case, `params[root_key] ||= {}` returns the original empty hash, which gets captured in `|h|`, but internally, a new empty `ActiveSupport::HashWithIndifferentAccess` is created and set as `params[root_key]`, which doesn't get modified.

## Reproduction

Here's a simple script to demonstrate this isolated issue:
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'activesupport'
end

require 'active_support/core_ext/hash/indifferent_access'
require 'active_support/version'

puts "Ruby version: #{RUBY_VERSION}"
puts "ActiveSupport version: #{ActiveSupport::VERSION::STRING}"
puts

p1 = {}
(p1[:a] ||= {})[:b] = 1
puts "With normal hash: #{p1}"

p2 = ActiveSupport::HashWithIndifferentAccess.new
(p2[:a] ||= {})[:b] = 1
puts "With HashWithIndifferentAccess: #{p2}"
```

**Output:**
```
Ruby version: 3.3.8
ActiveSupport version: 8.1.1

With normal hash: {:a=>{:b=>1}}
With HashWithIndifferentAccess: {"a"=>{}}
```

## Proposed Fix

Convert `params` to `to_h` before cloning.

ddnexus:
It is relevant because it should be usable by any frameworks, so it's a valid bug. Thank you for the report.

The ActiveSupport::HashWithIndifferentAccess concept and its impact in the ruby ecosystem is like Internet Explorer in Internet.

> Convert params to to_h before cloning.

Again, I forgot the evil entities that are messing up things behind my back! 🤣

ddnexus:
[43.1.6](https://github.com/ddnexus/pagy/releases/tag/43.1.6)

PR #851, opened 2025-12-13
Fix `root_key` nested params mutation in `compose_page_url` 

Pasniczy:
## Problem
Run into a bug when using the new `root_key` option. After page refresh on a URL with multiple nested keys like `/posts?articles[page]=1&news[page]=3` the pagination links were generated with `[page]=P ` (the `PAGE_TOKEN` placeholder) instead of the actual page numbers. 

## Root cause 
The `compose_page_url` method is using a shallow clone of request params:

```ruby
params = @request.params.clone(freeze: false)
```

so nested hashes (`params[root_key]`) still reference the original. When the method later modifies these nested hashes with `h[k] = v`, it is inadvertently modifying the original request params.

## Solution
Added deep cloning of nested params before modification:
```ruby
params[root_key] = params[root_key]&.then { |h| deep_dup_value(h) } if root_key
```

Now modifications only affect the local copy, preserving the original request params for subsequent operations.


ddnexus:
I implemented in a slightly different way, but the gist is the same!
Thank you!

Issue #858, opened 2026-01-12
Bug: anchor_string bug - option passed to pagy() not used by series_nav


andycroll:
### 👀 Before submitting...

- [x] I upgraded to pagy version 43.2.3
- [x] I asked [Pagy AI](https://gurubase.io/g/pagy) (Good starting point for further investigation).
- [x] I searched through the [Documentation](https://ddnexus.github.io/pagy/), [Issues](https://github.com/ddnexus/pagy/issues), and [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)

### 🧐 REQUIREMENTS

- [x] I am providing a VALID code file that confirms the bug
- [x] I am NOT posting any of the USELESS THINGS listed above
- [x] I am aware that this issue will be automatically closed if the code file is missing or INVALID

### 💬 Description

Pagy version: 43.2.3

## Description

The `anchor_string` option passed to the paginator (either via `Pagy.options` or per-paginator) is not picked up by the `a_lambda` method, so it doesn't appear in pagination links generated by `series_nav`.

## Expected behavior

Setting `anchor_string` at the paginator level should add that string to pagination links:

```ruby
# Controller
@pagy, @records = pagy(collection, anchor_string: 'data-turbo="true"')

# View - should output links with data-turbo="true"
<%== @pagy.series_nav %>
```

## Actual behavior

The anchor_string is stored in @options but never read. Links are rendered without the custom attributes.

## Root cause

In `lib/pagy/toolbox/helpers/support/a_lambda.rb`, the method signature defaults `anchor_string` to `nil` instead of reading from `@options`:

```ruby
def a_lambda(anchor_string: nil, **)  # <- defaults to nil
```

Other helper methods correctly read from `@options`. For example, in `support/series.rb`:

```ruby
def series(slots: @options[:slots] || SERIES_SLOTS, compact: @options[:compact], **)
```

## Suggested fix

Change the default to read from `@options[:anchor_string]`:

```ruby
def a_lambda(anchor_string: @options[:anchor_string], **)
```

This would make `anchor_string` consistent with other options like slots and compact.

## Workaround

I've temporarily added this patch in my initializer.

```ruby
Pagy.prepend(Module.new do
  protected

  def a_lambda(anchor_string: @options[:anchor_string], **)
    super
  end
end)
```

andycroll:
#### `repro.ru`

```ruby
# frozen_string_literal: true

# DESCRIPTION
[...]

PR #859, opened 2026-01-12
Fix anchor_string option not being read from @options in a_lambda

andycroll:
Fix for #858

When `anchor_string` was passed to `pagy()` it was stored in `@options` but ignored by `a_lambda` because the parameter defaulted to `nil` instead of reading from `@options`.

This is inconsistent with other methods like `series()` which correctly default their parameters from `@options` (e.g., `slots: @options[:slots]`).

This ensures that `anchor_string` attributes are included in generated links when passed to the `pagy()` method.

Hope this is useful.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -45,8 +45,6 @@ If you upgrade from version `< 9.0.0` see the following:
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


ddnexus:
Thank you!

PR #862, opened 2026-01-16
Add a SARGable predicate to keyset pagination

AlexMooney:
### Problem

The queries that keyset pagination constructs are not ideal for database index usage.

### Solution

Add an extra predicate to the pagination so that the query becomes SARGable.  Testing the queries on a postgres table with ~500,000,000 rows, three `order` keys, and a suitable index this change takes the query plan from timing out at 5 minutes down to `Planning Time: 67.137 ms Execution Time: 53.168 ms`.

I don't love the test checking the SQL strings; LMK if you have other ideas about how to test this.

### DIFF
--- a/gem/lib/pagy/classes/keyset/keyset.rb
+++ b/gem/lib/pagy/classes/keyset/keyset.rb
@@ -86,10 +86,26 @@ class Pagy
     end
 
     def fetch_records
+      apply_where(sargable_predicate, arguments_from(@prior_cutoff)) if @prior_cutoff && @keyset.size > 1
       apply_where(compose_predicate, arguments_from(@prior_cutoff)) if @prior_cutoff
       @set.limit(@limit + 1).to_a.tap do |records|
         @more = records.size > @limit && !records.pop.nil?
       end
+    end
+
+    # Create a SARGAable predicate from the first keyset column.
+    # For example, with a set like Pet.order(animal: :asc, name: :desc, id: :asc)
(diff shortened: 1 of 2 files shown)


ddnexus:
@AlexMooney Thanks for the PR.

I am surprised by your results. It obviously does a full scan... when it should not.

Please, check this for me:

- I guess that the "suitable index" is a composite **B-tree** type index that matches the ORDER BY columns **and** their directions (ASC/DESC) exactly as the `@keyset`. Please confirm.
- Just to avoid obscure interactions, try it without any other conditions (i.e. the only query is the composite predicate one) 
- What DB/version engine are you using?

If all the above is OK and the version is modern, then relational DBs are still not very clever. 🤷🏻‍♂️

BTW I just pushed a much simpler [implementation](https://github.com/ddnexus/pagy/commit/283ad4d22d7dcaff6b77d82b56bf7b7a4be1f8cb) to `dev`, and while at it I also made my previous code less verbose, simpler and more readable. 

Please, use the `dev` branch and let me know whether it fixes the problem. Thank you.

ddnexus:
Thank you.

Released in [43.2.7](https://github.com/ddnexus/pagy/releases/tag/43.2.7)

AlexMooney:
Hey sorry for disappearing there. I was on a long weekend vacation and was intending on getting you some queries for comparison tomorrow. Thanks for shipping the improvement.

PR #863, opened 2026-01-22
Fix keyset error caused by the change of resolve_page

ippachi:
Thank your for your great gem.

[This commit](https://github.com/ddnexus/pagy/commit/f859166c8cbb9d651ffb091e00337a501116c657) raises exception on keyset pagination.

It changes `@options[:page]` from `nil` to `''`, so JSON.parse raises exception.

> [!TIP]
> See also [Contributing](https://github.com/ddnexus/pagy/blob/master/.github/CONTRIBUTING.md) for details.

### DIFF
--- a/gem/lib/pagy/classes/keyset/keyset.rb
+++ b/gem/lib/pagy/classes/keyset/keyset.rb
@@ -82,7 +82,7 @@ class Pagy
     def keyset? = true
 
     def assign_page
-      return unless (@page = @options[:page])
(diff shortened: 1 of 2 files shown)


ddnexus:
Good catch. Thank you.

It should actually be a concern of the `resolve_page`. The policy is: strict internal implementation classes (require strict usage), but flexible paginators (ease the life of developers).

Implemented in e32bd0140fee855e0ab69faeb1fb26fa5f230d3c. You can try it with `dev`.

PR #873, opened 2026-02-12
Add test and fix empty aria label issue

mikepayready:
Starting in version 43.2.7, the aria label is blank when it is not explicitly overridden. I tracked it down to [this change](https://github.com/ddnexus/pagy/commit/eb7607df605ec359a5f282a9dc8b12184d66f511#diff-7ddc38c99a8b2ddd9a06b4942e648e88a0f6bbe21f86d00f336140b67a158632R8) which moves the default from within the method to the keyword argument default.

However, uses of this method [such as here](https://github.com/ddnexus/pagy/blob/eb7607df605ec359a5f282a9dc8b12184d66f511/gem/lib/pagy/toolbox/helpers/support/wrap_series_nav.rb#L16) explicitly pass a value that might be nil. Thus, the nil value overrides the keyword argument default and the aria label is empty.

### DIFF
--- a/gem/lib/pagy/toolbox/helpers/support/nav_aria_label_attribute.rb
+++ b/gem/lib/pagy/toolbox/helpers/support/nav_aria_label_attribute.rb
@@ -5,7 +5,8 @@ class Pagy
   private
 
   # Compose the aria label attribute for the nav
-  def nav_aria_label_attribute(aria_label: I18n.translate('pagy.aria_label.nav', count: @last))
+  def nav_aria_label_attribute(aria_label: nil)
+    aria_label ||= I18n.translate('pagy.aria_label.nav', count: @last)
     %(aria-label="#{aria_label}")
   end
 end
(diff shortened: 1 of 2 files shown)


mikepayready:
@ddnexus I wasn't sure if merging against your dev branch was what you would like, but I just rebased my branch onto that to resolve some conflicts. If you would like me to switch the target or anything just let me know 🙂

I hope you don't mind I went straight for a PR rather than opening an issue.

ddnexus:
Yes, sorry for the radio silence...

Lots of stuff going on at the same time:

- moving to another country
- porting my main laptop from ubuntu to fedora
- my IDE from RubyMine to VScode
- my dev env from docker-compose to podman pod
- projects from ruby 3.* to 4.*
- pagy e2e testing from cypress to ferrum
- squashing bugs

Luckily I have no plans for changing wife, sex, or pronouns 😅

Now... what was the topic again? 

Thanks for the PR: usually dev is the right choice, but since you are available to rebase I will let you know very soon...

Thank you

mikepayready:
@ddnexus no apologies needed, appreciate the awesome gem and your time!

After updating Pagy to the latest version, I had a test failure in my Rails app where I was asserting the aria label text to be present, but it was missing.

I tracked it down to the change I pointed to in my previous comment, and added a test that fails because of the same issue along with the fix (reverting back to the code that was working).

Happy to make any adjustments you would like, or help out if you would like a different fix. I will be away until next Tuesday though, but happy to tackle any changes you would like when I'm back.

Please no pressure intended to release a fix though. Just happy to help get a fix in 🙂

PR #874, opened 2026-02-13
Fix missing bracket in series_nav_js documentation

KieraninTehdas:
Surely the pinnacle of my career. Adds a missing bracket in the docs for series_nav_js which caught me out when I blindly copy pasted the example :flushed:


ddnexus:
Hello fellow "Professional side project starter"! We are colleagues! 🤭
Thank you for the fix. Looking forward to see it in your CV. 🤓

PR #893, opened 2026-04-18
Fix typo in docs

mcauser:
`Rails.env.development?` instead of `Rails.development?` on https://ddnexus.github.io/pagy/resources/javascript/


ddnexus:
@mcauser Thank you!

PR #895, opened 2026-04-21
Fix #page_url duplicating query params when page_key is a symbol

### LINKED ISSUE #848: Bug: #page_url duplicates existing query string parameters
### 👀 Before submitting...

- [x] I upgraded to pagy version 43.2.0
- [x] I asked [Pagy AI](https://gurubase.io/g/pagy) (Good starting point for further investigation).
- [x] I searched through the [Documentation](https://ddnexus.github.io/pagy/), [Issues](https://github.com/ddnexus/pagy/issues), and [Q&A](https://github.com/ddnexus/pagy/discussions/categories/q-a)

### 🧐 REQUIREMENTS

- [x] I am providing a VALID code file that confirms the bug
- [x] I am NOT posting any of the USELESS THINGS listed above
- [x] I am aware that this issue will be automatically closed if the code file is missing or INVALID

### 💬 Description

If the current URL is:

http://localhost:3000/users/1?page=1

@pagy.page_url(2) outputs the following URL:

http://localhost:3000/users/1?page=1&page=2

This is incorrect. Any existing query string keys should be replaced, not duplicated.

The expected output in the provided example would be:

http://localhost:3000/users/1?page=2

alexander-makarenko:
`Pagy::Request#get_params` stores request params as a plain string-keyed hash. A symbol `page_key` / `limit_key` passed through options would write a sibling symbol entry alongside the existing string key, emitting duplicate pairs in the query string. Coerce both keys to string in `Linkable#compose_page_url` so the container is always updated in place.

Closes #848.

### DIFF
--- a/gem/lib/pagy/modules/abilities/linkable.rb
+++ b/gem/lib/pagy/modules/abilities/linkable.rb
@@ -52,8 +52,8 @@ class Pagy
                     params
                   end
 
-      { opts[:page_key]  => compose_page_param(page),
-        opts[:limit_key] => opts[:max_limit] && opts[:limit] }.each do |k, v|
+      { opts[:page_key].to_s  => compose_page_param(page),
+        opts[:limit_key].to_s => opts[:max_limit] && opts[:limit] }.each do |k, v|
         v ? container[k] = v : container.delete(k)
       end
 
--- a/test/unit/pagy/modules/abilities/linkable_test.rb
+++ b/test/unit/pagy/modules/abilities/linkable_test.rb
@@ -148,5 +148,19 @@ describe 'Pagy::Linkable Specs' do
       url = subject.call_compose_page_url(2)
       _(url).must_equal_url '/foo?limit=10&page=2'
     end
+
+    it 'does not duplicate params when page_key is a symbol' do
+      subject = linkable_class.new(request_params: { 'page' => '1' },
+                                   options:        { page_key: :page })
+      url = subject.call_compose_page_url(2)
+      _(url).must_equal_url '/foo?page=2'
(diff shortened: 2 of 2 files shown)


ddnexus:
Thank you for this PR.

The very reason they are strings is that urls (keys and params) are [always strings](https://ddnexus.github.io/pagy/resources/urls/#params), and we don't coerce anything on purpose.

It is not an expectation, so they don't need a fix. Maybe validating the value type at initialization and raise an option error if they fail could ensure more strictness.

alexander-makarenko:
@ddnexus I think the current behaviour is surprising.

Wouldn't it more intuitive and convenient to have both strings and symbols accepted and treated equally?
This pattern is quite common and idiomatic in Ruby and especially in Rails. Would there be any harm in adopting the same approach here?

If there's a specific reason why symbols are not supported, one would expect the method to raise rather than silently accept the option and produce a result that's clearly incorrect.

ddnexus:
> Would there be any harm in adopting the same approach here?

Yes, inconsistencies in the code which tries to follow rules to keep it clean and efficient. In this particular case it would be minimal, however... it would hurt my feelings. 🫨🤓

> If there's a specific reason why symbols are not supported, one would expect the method to raise rather than silently accept the option and produce a result that's clearly incorrect.

Yes, that's what I was saying. The current behavior is indeed an inconsistency itself. The other options are validated at initialize time, or they will generate an error somewhere anyway. Instead, these two are not validated and don't generate any error, so raising is the right behavior to implement.

Will do.

ddnexus:
Implemented in the dev branch

PR #904, opened 2026-05-18
Add a real rake test task for the unit suite

ShyamBorsaniya:
- The root [Rakefile:1-12] previously only loaded the Thor tasks, so [rake test] completed without running any tests.

- Added Bundler setup and a dedicated [test] task that executes the full unit suite under [gem/lib:test].

- Set [test] as the default Rake task so the conventional command now behaves as expected.
- Verified the change by running [rake test] the full suite passes with 562 tests and 0 failures.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,13 @@
 # frozen_string_literal: true
 
+require 'bundler/setup'
+
 Rake.add_rakelib 'tasks'
+
+desc 'Run the unit test suite'
+task :test do
+  script = 'Dir.glob("test/unit/**/*_test.rb").sort.each { |f| require "./" + f }'
+  sh Gem.ruby, '-Igem/lib:test', '-e', script
+end
+
+task default: :test


ddnexus:
Thank you for your PR!

However `Dir.glob("test/unit/**/*_test.rb").sort.each { |f| require "./" + f }` does not actually "test" pagy. It only runs all tests files in sorted order, which - when running that way - covers only a quite reduced subset of the cases we must cover.

[This](https://github.com/ddnexus/pagy/blob/master/tasks/api.thor) is what actually ensures that the API _fully_ works. And [this](https://github.com/ddnexus/pagy/blob/master/tasks/test.thor) ensures that the API is 100% covered, that the E2E is fine on the client side, and that the code is following the best practices and the pagy standards (rubocop).

Preparing a `rake test`, _(covering all we need, the way we need it)_ would be quite cumbersome and painful. That's why we use Thor, which simplify the code and maintenance. You can run `thor test:all` to actually run what you would expect the `rake task` would run. _(see this [README](https://github.com/ddnexus/pagy/tree/master/test) for details)_

<details>
<summary><code>thor test:all</code> output</summary>
<br>
As you can see the units that your <code>rake test</code> run are roughly half of what we need to run (just for the API):
<p>

```
/usr/bin/zsh -c "bundle exec thor test:all"
Inspecting 200 files
.....................................................................................................
...................................................................................................

200 files inspected, no offenses detected

Testing the CURRENT API...
Started with run options --seed 48225

  570/570: [====================================================] 100% Time: 00:00:00, Time: 00:00:00

Finished in 0.54893s
570 tests, 1667 assertions, 0 failures, 0 errors, 0 skips

Testing the NEXT API...
Started with run options --seed 36839

  560/560: [====================================================] 100% Time: 00:00:00, Time: 00:00:00

Finished in 0.52270s
560 tests, 1628 assertions, 0 failures, 0 errors, 0 skips

Coverage report generated for CURRENT_API, NEXT_API to /home/dd/repos/dd/pagy/coverage.
Line Coverage: 100.0% (1414 / 1414)
Branch Coverage: 100.0% (399 / 399)

Testing E2E apps...
Started with run options --seed 64863

  19/19: [======================================================] 100% Time: 00:00:13, Time: 00:00:13

Finished in 13.24284s
19 tests, 1217 assertions, 0 failures, 0 errors, 0 skips

Process finished with exit code 0
```

</details>

HTH

Thank you again for your interest in improving pagy!

Issue #913, opened 2026-07-17
Docs: 43 upgrade guide 404s


righteous-oak:
### Description

The [upgrade guide to version 43](https://github.com/ddnexus/pagy/blob/master/guides/upgrade-guide), as linked to in the [changelog](https://github.com/ddnexus/pagy/blob/master/docs/CHANGELOG.md), 404s.

benkoshy:
### The problem

- Links from the github repo 404.
- Links when when read from the docs site work fine i.e. https://ddnexus.github.io/pagy/changelog/

### Expectation
The expectation right now is that people will be reading the change logs from the documentation site here:  https://ddnexus.github.io/pagy/changelog/ and accordingly, from that location the links should work.  

### Potential Solution

The solution is really simple: use relative paths.

```diff
-Follow the [Upgrade to 43 Guide](/guides/upgrade-guide).
+Follow the [Upgrade to 43 Guide](./guides/upgrade-guide.md).
```
* That should work from both Github and Retype.
* We must add the `.md` extension for it to work in both locations.

Thank you @righteous-oak .

benkoshy:
~~There are a few other links like this within the docs that could be amended. A simple regex / FIND will uncover them. If anyone is interested, please have at it via PR:~~

~~e.g. here is a regex that leads to some results. Naturally you will want to be selective on which links to update:~~

```sh
\]\((?! \.)
```

~~This will uncover any link that is NOT a relative path link. You can update the regex to make the task simpler.~~

ddnexus:
@ben, retype handles the change you did in the #917 quite well. However using relative paths will link the relative file, but then we will have the user navigating all the docs in github while formatted for retype, which is not what we want.

The CHANGELOG, CHANGELOG_LEGACY and README are the only files that should be readable also in github, because conventionally they are often viewed in github. For that reason, their links should work from github.

However, all their links pointing to the OTHER files in the docs, should be reached with an absolute url. That way that documents will be ok read from github, but their links to the docs will point to the docs URL.

I believe the README already follows that rule, while the CHANGELOG file do not, and they should.

Does all that make sense?

benkoshy:
> [@ben](https://github.com/ben), retype handles the change you did in the [#917](https://github.com/ddnexus/pagy/pull/917) quite well. However using relative paths will link the relative file, but then we will have the user navigating all the docs in github while formatted for retype, which is not what we want.

[...]

PR #918, opened 2026-07-27
Chore: url override to github actions

benkoshy:
Why?

So anyone who contributes to the docs can test it "locally", with their own deployed version on their own github url e.g. benkoshy.githbub.io/pagy, prior to making a PR to the upstream repository.

The genesis was issue https://github.com/ddnexus/pagy/issues/913 I wanted to test a change to the docs on my own deployed version: benkoshy.github.io/pagy.

I could not do so unless i set the retype URL to my own account benkoshy.github.io. But doing so would affect the deployment at ddnexus.github.io.

The solution is to override the url, which allows the benkoshy built docs to work  without affecting ddexus.github.io/pagy

Check out my docs here: https://benkoshy.github.io/pagy/

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -43,6 +43,7 @@ jobs:
           RETYPE_KEY: ${{ secrets.RETYPE_KEY }}
         with:
           config_path: docs/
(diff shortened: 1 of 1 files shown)


ddnexus:
I guess this does not affect the local development at all, so it is a very clever way to solve the problem.

PR #920, opened 2026-08-23
Fix nil records returned for Array collections on page overflow

### LINKED ISSUE #919: Bug: Array pagination returns nil records (not []) on page overflow
### Disclaimer

- [x] I am aware that this issue will be closed or converted to a discussion if the requested code is missing or invalid.

### Description

With the default range handling (`raise_range_error: false`, i.e. the "empty page" mode that replaced the old `overflow: :empty_page` extra), paginating a plain `Array` with an out-of-range page returns `records = nil` instead of an empty array.

The `Pagy::Offset` object itself is correctly built in empty-page mode (`pagy.in == 0`, `pagy.previous == pagy.last`), but `OffsetPaginator.paginate` extracts the records with:

```ruby
records = if collection.instance_of?(Array)
            collection[pagy.offset, pagy.limit]   # [1, 2, 3][16, 2] => nil, not []
          else
            pagy.records(collection)              # offset().limit() => empty relation
          end
```

Ruby's `Array#[](start, length)` returns `nil` when `start > size`, so the two branches disagree: ActiveRecord (and any offset/limit collection) honors the empty-page contract by returning an empty collection, while `Array` returns `nil`. Any subsequent `records.map` / `records.each` in the app then raises `NoMethodError: undefined method 'map' for nil`.

This is easy to hit in production with stale URLs carrying an overflowing `page` param (bookmarks, browser history, prefetch, or a filter change that shrinks the result set while the page param sticks).

### Reproduction (single file)

```ruby
# frozen_string_literal: true
# ruby pagy_issue.rb  (pagy 43.6.1)

require 'pagy'

class App
  include Pagy::Method
  public :pagy
end

app        = App.new
collection = [1, 2, 3]

# In-range page: OK
_, records = app.pagy(:offset, collection, limit: 2, request: { params: { 'page' => '2' } })
puts "page 2 (in range):  records = #{records.inspect}"   # => [3]

# Overflow page: pagy is in empty-page mode, but records is nil instead of []
pagy, records = app.pagy(:offset, collection, limit: 2, request: { params: { 'page' => '9' } })
puts "page 9 (overflow):  pagy.in = #{pagy.in}, records = #{records.inspect}"

records.map { |r| r } # NoMethodError: undefined method 'map' for nil
```

Output (pagy 43.6.1, ruby 4.0.6):

```
page 2 (in range):  records = [3]
page 9 (overflow):  pagy.in = 0, records = nil
pagy_issue.rb:22:in '<main>': undefined method 'map' for nil (NoMethodError)
```

### Expected

`records == []` on overflow, consistent with non-Array collections and with the documented empty-page behavior. One-character fix in `gem/lib/pagy/toolbox/paginators/offset.rb`:

```ruby
collection[pagy.offset, pagy.limit] || []
```

Happy to send a PR with the fix and a test if you want.

As a side note: it would be great if overflow handling stayed fully native like this across all collection types, the way the old overflow extra made it a non-issue — today every app paginating Arrays has to wrap the result defensively to survive stale page params.

leoplct:
Fixes #919

### Problem

With the default `raise_range_error: false` (empty-page mode), `OffsetPaginator.paginate` extracts Array records with `collection[pagy.offset, pagy.limit]`. Ruby's `Array#[](start, length)` returns `nil` (not `[]`) when `start > size`, so an out-of-range page returns `records = nil` for Array collections, while every other collection type returns an empty collection (`offset().limit()` → empty relation). Any subsequent `records.map` / `records.each` in the app raises `NoMethodError: undefined method 'map' for nil` — easy to hit in production with stale URLs carrying an overflowing `page` param (bookmarks, history, prefetch, or a filter change that shrinks the result set).

### Fix

Normalize the Array slice with `|| []`, restoring the empty-page contract natively for Arrays too, consistent with the other branches.

### Test

Added a unit test in `test/unit/pagy/toolbox/paginators/offset_test.rb` (`with Array` section): an overflowing page must return `pagy.in == 0` and `records == []`. It fails with `NoMethodError`-inducing `nil` before the fix and passes after. Full paginator unit test suite and rubocop pass locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -29,19 +29,19 @@ Increment the MAJOR version in your Gemfile ONLY when you are ready to handle th
 
 ### Breaking Changes
 
-Follow the [Upgrade to 43 Guide](/guides/upgrade-guide).
+Follow the [Upgrade to 43 Guide](https://ddnexus.github.io/pagy/guides/upgrade-guide).
 
 If you upgrade from version `< 9.0.0` see the following:
 
-- [Breaking changes in version 9.0.0](/CHANGELOG_LEGACY.md#version-900)
-- [Breaking changes in version 8.0.0](/CHANGELOG_LEGACY.md#version-800)
-- [Breaking changes in version 7.0.0](/CHANGELOG_LEGACY.md#version-700)
-- [Breaking changes in version 6.0.0](/CHANGELOG_LEGACY.md#version-600)
-- [Breaking changes in version 5.0.0](/CHANGELOG_LEGACY.md#version-500)
-- [Breaking changes in version 4.0.0](/CHANGELOG_LEGACY.md#version-400)
-- [Breaking changes in version 3.0.0](/CHANGELOG_LEGACY.md#version-300)
-- [Breaking changes in version 2.0.0](/CHANGELOG_LEGACY.md#version-200)
-- [Breaking changes in version 1.0.0](/CHANGELOG_LEGACY.md#version-100)
+- [Breaking changes in version 9.0.0](https://ddnexus.github.io/pagy/changelog_legacy#version-900)
+- [Breaking changes in version 8.0.0](https://ddnexus.github.io/pagy/changelog_legacy#version-800)
+- [Breaking changes in version 7.0.0](https://ddnexus.github.io/pagy/changelog_legacy#version-700)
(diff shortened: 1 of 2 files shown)


ddnexus:
Thank you!

